PR #73, opened 2023-11-23
Update readme to use github.com URLs rather than relative URLs

muddyfish:
## Description

Update readme to use github.com URLs rather than relative URLs

## Additional context

On sites like PyPI we want to link people to github rather than keep them on-site.

## Testing

This was the Readme we pushed to PyPI

--------


muddyfish:
Closing PR to re-open in a fork

PR #86, opened 2023-12-01
Align imports and fix docs

dnanuti:
Imports for S3Reader and S3Writer are not consistent.
Update their imports. Align the order they show up on API reference page.

## Description

Enhancements for our API docs: https://awslabs.github.io/s3-connector-for-pytorch/

## Additional context

* No breaking change. 

## Testing

`sphinx-build -b html . _build/html`

--------


s3torchconnector/src/s3torchconnector/_s3client/_s3client.py
```diff
@@ -5,9 +5,7 @@
 from functools import partial
 from typing import Optional, Any
 
-from .s3reader import S3Reader
-from .s3writer import S3Writer
-
+from s3torchconnector import S3Reader, S3Writer
```
fuatbasik, line 8: do we use s3torchconnector or .?

fuatbasik (dismissed):
Thanks. Just 1 nitpick but lgtm otherwise.

muddyfish (dismissed):

s3torchconnector/src/s3torchconnector/_s3client/_s3client.py
dnanuti: They are moved to the main module now, that's why we use `s3torchconnector`.

fuatbasik (approved):

PR #87, opened 2023-12-05
Avoid excessive memory consumption when utilizing s3map_dataset 

IsaevIlya:
Fix for https://github.com/awslabs/s3-connector-for-pytorch/issues/89

## Description
The goal of the changes was to minimize the retention time for S3 data loaded in memory. This was achieved by delaying the creation of `S3Readers` until right before reading the given object. Immediately after reading finishes, ownership of the `S3Reader` containing the loaded content is transferred to the calling client.

## Testing
Integration and unit test were run.

--------

### DIFF
--- a/s3torchconnector/src/s3torchconnector/_s3bucket_key.py
+++ b/s3torchconnector/src/s3torchconnector/_s3bucket_key.py
@@ -1,23 +1,10 @@
 #  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #  // SPDX-License-Identifier: BSD
+from typing import NamedTuple
 
-class S3BucketKey(object):
+
+class S3BucketKey(NamedTuple):
     """Read-only information about object stored in S3."""
 
-    def __init__(
-            self,
-            bucket: str,
-            key: str,
-    ):
-        if not bucket:
-            raise ValueError("Bucket should be specified")
-        self._bucket = bucket
-        self._key = key
-
-    @property
-    def bucket(self):
-        return self._bucket
-
-    @property
-    def key(self):
-        return self._key
+    bucket: str
+    key: str
--- a/s3torchconnector/src/s3torchconnector/_s3bucket_key.py
+++ b/s3torchconnector/src/s3torchconnector/_s3bucket_key.py
@@ -1,0 +1,23 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  // SPDX-License-Identifier: BSD
+
+class S3BucketKey(object):
+    """Read-only information about object stored in S3."""
+
+    def __init__(
+            self,
+            bucket: str,
+            key: str,
+    ):
+        if not bucket:
+            raise ValueError("Bucket should be specified")
+        self._bucket = bucket
+        self._key = key
+
+    @property
+    def bucket(self):
+        return self._bucket
+
+    @property
+    def key(self):
+        return self._key
--- a/s3torchconnector/src/s3torchconnector/_s3dataset_common.py
+++ b/s3torchconnector/src/s3torchconnector/_s3dataset_common.py
@@ -1,7 +1,7 @@
 #  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #  // SPDX-License-Identifier: BSD
 
-from typing import Iterable, Union, Tuple, List
+from typing import Iterable, Union, Tuple
 
 from ._s3_bucket_iterable import S3BucketIterable
 from ._s3client import S3Client
@@ -44,15 +44,10 @@ def get_objects_from_uris(
     # TODO: We should be consistent with URIs parsing. Revise if we want to do this upfront or lazily.
     bucket_key_pairs = [parse_s3_uri(uri) for uri in object_uris]
 
-    return bucket_key_pairs_to_objects(bucket_key_pairs)
+    return (S3BucketKey(bucket, key) for bucket, key in bucket_key_pairs)
 
 
-def bucket_key_pairs_to_objects(bucket_key_pairs: List[Tuple[str, str]]):
-    for bucket, key in bucket_key_pairs:
-        yield S3BucketKey(bucket, key)
-
-
-def list_objects_from_prefix(s3_uri: str, client: S3Client) -> Iterable[S3BucketKey]:
+def get_objects_from_prefix(s3_uri: str, client: S3Client) -> Iterable[S3BucketKey]:
     bucket, prefix = parse_s3_uri(s3_uri)
     s3objects = S3BucketIterable(client, bucket, prefix)
     bucket_key_pairs = (S3BucketKey(obj.bucket, obj.key) for obj in s3objects)
--- a/s3torchconnector/src/s3torchconnector/_s3dataset_common.py
+++ b/s3torchconnector/src/s3torchconnector/_s3dataset_common.py
@@ -41,8 +41,7 @@ def get_objects_from_uris(
 ) -> Iterable[S3BucketKey]:
     if isinstance(object_uris, str):
         object_uris = [object_uris]
-    # TODO: We should be consistent with URIs parsing. Revise if we want to do this upfront or lazily.
-    bucket_key_pairs = [S3BucketKey(*parse_s3_uri(uri)) for uri in object_uris]
+    bucket_key_pairs = (S3BucketKey(*parse_s3_uri(uri)) for uri in object_uris)
 
     return bucket_key_pairs
 
--- a/s3torchconnector/src/s3torchconnector/_s3dataset_common.py
+++ b/s3torchconnector/src/s3torchconnector/_s3dataset_common.py
@@ -1,7 +1,7 @@
 #  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #  // SPDX-License-Identifier: BSD
 
-from typing import Iterable, Union, Tuple
+from typing import Iterable, Union, Tuple, List
 
 from ._s3_bucket_iterable import S3BucketIterable
 from ._s3client import S3Client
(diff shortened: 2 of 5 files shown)


s3torchconnector/src/s3torchconnector/_s3bucket_key.py
```diff
@@ -0,0 +1,23 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  // SPDX-License-Identifier: BSD
+
+class S3BucketKey(object):
```
muddyfish, line 4: `(object)` is not needed.

muddyfish, line 4: A `dataclass` or `namedtuple` would work better here.

s3torchconnector/src/s3torchconnector/_s3dataset_common.py
```diff
@@ -41,22 +41,18 @@ def parse_s3_uri(uri: str) -> Tuple[str, str]:
...
-):
-    for bucket, key in bucket_key_pairs:
-        yield client.get_object(bucket, key)
-
-
-def list_objects_from_prefix(s3_uri: str, client: S3Client) -> S3BucketIterable:
+def list_objects_from_prefix(s3_uri: str, client: S3Client) -> Iterable[S3BucketKey]:
     bucket, prefix = parse_s3_uri(s3_uri)
-    return S3BucketIterable(client, bucket, prefix)
+    s3objects = list(S3BucketIterable(client, bucket, prefix))
```
muddyfish, line 56: I think this is the only place we're using `S3BucketIterable`. We should be able to completely refactor this out and just use `client.list_objects` here. Might be a bit much for this PR though

s3torchconnector/src/s3torchconnector/_s3dataset_common.py
```diff
@@ -41,22 +41,18 @@ def parse_s3_uri(uri: str) -> Tuple[str, str]:
...
+    bucket_key_pairs = [S3BucketKey(obj.bucket, obj.key) for obj in s3objects]
```
muddyfish, line 57: We were deliberately returning an iterable here rather than a list. This uses less memory

s3torchconnector/src/s3torchconnector/s3iterable_dataset.py
```diff
@@ -91,5 +91,8 @@ def _get_client(self):
             self._client = S3Client(self.region)
         return self._client
 
+    def _get_transformed_object(self, bucket_key) -> S3Reader:
```
muddyfish, line 94: Nit: Add type hint for `bucket_key`.

s3torchconnector/src/s3torchconnector/_s3bucket_key.py
fuatbasik: Do we really need a dataclass or namedtuple here? Can we just use tuples as (bucket,key)? What is the convention for such simple encapsulations?

s3torchconnector/src/s3torchconnector/_s3dataset_common.py
fuatbasik: +1 but I would leave it as a follow-up PR.

s3torchconnector/src/s3torchconnector/_s3bucket_key.py
IsaevIlya: Intention was to save visibility on what we are storing there. I'm afraid, tuple is not providing that.

s3torchconnector/src/s3torchconnector/_s3bucket_key.py
```diff
@@ -0,0 +1,23 @@
...
+            bucket: str,
+            key: str,
+    ):
+        if not bucket:
+            raise ValueError("Bucket should be specified")
+        self._bucket = bucket
+        self._key = key
+
+    @property
+    def bucket(self):
```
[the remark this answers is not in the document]
IsaevIlya: I and @muddyfish discussed, that it would be better to use `namedtuple` instead of this class.

s3torchconnector/src/s3torchconnector/_s3bucket_key.py
muddyfish: As far as speed is concerned, `namedtuple` performs the same as regular `tuple`. 
It also provides simple getters for the properties so we can do `bucket_key.key` rather than `bucket_key[1]`.

s3torchconnector/src/s3torchconnector/_s3dataset_common.py
fuatbasik: If we start returning a list here, would it eagerly fetch all the keys for IterableDataset? IterableDataset should fetch keys from S3 as we iterate over records. When we are returning an iterable thats the behaviour. If we convert it to list, we might change this behaviour.

s3torchconnector/src/s3torchconnector/s3map_dataset.py
```diff
@@ -32,19 +33,17 @@ def __init__(
...
     @property
-    def _dataset_objects(self) -> List[S3Reader]:
-        if self._dataset_object_store is None:
-            self._dataset_object_store = list(
-                self._get_dataset_objects(self._get_client())
-            )
-        return self._dataset_object_store
+    def _dataset_bucket_key_pairs(self) -> List[S3BucketKey]:
+        if self._bucket_key_pairs is None:
+            self._bucket_key_pairs = list(self._get_dataset_objects(self._get_client()))
```
muddyfish, line 45: We've changed the return type of `self._get_dataset_objects`, so also need to change the type in the definition of `__init__`

s3torchconnector/src/s3torchconnector/s3map_dataset.py
```diff
@@ -102,8 +101,12 @@ def _get_client(self):
             self._client = S3Client(self.region)
         return self._client
 
+    def _get_object(self, i) -> S3Reader:
```
muddyfish, line 104: Nit: typehint `i` and perhaps give it a better name

s3torchconnector/src/s3torchconnector/_s3dataset_common.py
```diff
@@ -42,22 +38,18 @@ def parse_s3_uri(uri: str) -> Tuple[str, str]:
 
 def get_objects_from_uris(
     object_uris: Union[str, Iterable[str]], client: S3Client
-) -> Iterable[S3Reader]:
+) -> Iterable[S3BucketKey]:
     if isinstance(object_uris, str):
         object_uris = [object_uris]
     # TODO: We should be consistent with URIs parsing. Revise if we want to do this upfront or lazily.
-    bucket_key_pairs = [parse_s3_uri(uri) for uri in object_uris]
+    bucket_key_pairs = [S3BucketKey(*parse_s3_uri(uri)) for uri in object_uris]
```
fuatbasik, line 45: This is unintentionally changing the behaviour of this method. This was supposed to return an iterable but currently returning a list. The issue is not with this PR but actual implementation in the first place. Lets have another look at this part.

s3torchconnector/src/s3torchconnector/_s3dataset_common.py
```diff
@@ -42,22 +38,18 @@ def parse_s3_uri(uri: str) -> Tuple[str, str]:
...
-
-
-def list_objects_from_prefix(s3_uri: str, client: S3Client) -> S3BucketIterable:
+def list_objects_from_prefix(s3_uri: str, client: S3Client) -> Iterable[S3BucketKey]:
     bucket, prefix = parse_s3_uri(s3_uri)
-    return S3BucketIterable(client, bucket, prefix)
+    s3objects = S3BucketIterable(client, bucket, prefix)
+    bucket_key_pairs = (S3BucketKey(obj.bucket, obj.key) for obj in s3objects)
+
+    return bucket_key_pairs
```
fuatbasik, line 55: This is unintentionally changing the behaviour of this method. This was supposed to return an iterable but currently returning a list. The issue is not with this PR but actual implementation in the first place. Lets have another look at this part. This would change the logic of S3IterableDataset.

fuatbasik (changes requested):

muddyfish (approved):

s3torchconnector/src/s3torchconnector/_s3dataset_common.py
```diff
@@ -42,22 +38,23 @@ def parse_s3_uri(uri: str) -> Tuple[str, str]:
...
     object_uris: Union[str, Iterable[str]], client: S3Client
-) -> Iterable[S3Reader]:
+) -> Iterable[S3BucketKey]:
     if isinstance(object_uris, str):
         object_uris = [object_uris]
     # TODO: We should be consistent with URIs parsing. Revise if we want to do this upfront or lazily.
     bucket_key_pairs = [parse_s3_uri(uri) for uri in object_uris]
 
-    return bucket_key_pairs_to_objects(bucket_key_pairs, client)
+    return bucket_key_pairs_to_objects(bucket_key_pairs)
```
muddyfish, line 47: Nit: I'd prefer `return (S3BucketKey(bucket, key) for bucket, key in bucket_key_pairs)` here but what we've got is fine.
Or even just returning an actual list instead would be fine

s3torchconnector/tst/unit/test_s3dataset_common.py
```diff
@@ -69,16 +69,13 @@ def test_list_objects_from_prefix(
     prefix: str, keys: Sequence[str], expected_count: int
 ):
     mock_client = _create_mock_client_with_dummy_objects(TEST_BUCKET, keys)
-    objects = list_objects_from_prefix(f"{S3_PREFIX}/{prefix}", mock_client)
+    bucket_key_pairs = list_objects_from_prefix(f"{S3_PREFIX}/{prefix}", mock_client)
     count = 0
-    for index, object in enumerate(objects):
+    for index, bucket_key_pair in enumerate(bucket_key_pairs):
         count += 1
```
dnanuti, line 75: Why do we still have count? Shouldn't index be enough?

muddyfish (dismissed):

s3torchconnector/src/s3torchconnector/_s3dataset_common.py
dnanuti: Do we want a list? Didn't we want them lazily loaded?

s3torchconnector/src/s3torchconnector/_s3dataset_common.py
```diff
@@ -42,22 +38,23 @@ def parse_s3_uri(uri: str) -> Tuple[str, str]:
...
 
 
-def bucket_key_pairs_to_objects(
-    bucket_key_pairs: List[Tuple[str, str]], client: S3Client
-):
+def bucket_key_pairs_to_objects(bucket_key_pairs: List[Tuple[str, str]]):
```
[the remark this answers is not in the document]
dnanuti: I would prefer to drop this and use the generator above.

fuatbasik (approved):
Put some non-blocking comments.

s3torchconnector/tst/unit/test_s3dataset_common.py
IsaevIlya: I think, we need `count` to prevent situation when number of returned objects will be smaller than number of `expected_keys`. We use `count` in assert below the "for" loop.

dnanuti (approved):

muddyfish (approved):

PR #88, opened 2023-12-05
Add a 2-minute timeout to tests

muddyfish:
## Description

Previously we were having issues where our tests would take too long. Now they're set to time out after a specified period.

## Testing

Tests pass.

--------

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -10,8 +10,10 @@ from hypothesis import settings
 settings.register_profile("ci", max_examples=1000, deadline=timedelta(seconds=1))
 settings.load_profile(os.getenv("HYPOTHESIS_PROFILE", "default"))
 
+per_test_timeout = int(os.getenv("TEST_TIMEOUT", "120"))
+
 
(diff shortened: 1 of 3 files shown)


conftest.py
```diff
@@ -4,7 +4,16 @@
...
 settings.register_profile("ci", max_examples=1000, deadline=timedelta(seconds=1))
 settings.load_profile(os.getenv("HYPOTHESIS_PROFILE", "default"))
+
+per_test_timeout = int(os.getenv("TEST_TIMEOUT", "120"))
+
+
+def pytest_collection_modifyitems(items):
+    for item in items:
+        if item.get_closest_marker("timeout") is None:
+            item.add_marker(pytest.mark.timeout(per_test_timeout))
```
dnanuti, line 19: So this still allows use to specify a different timeout for a particular test, right?

dnanuti (approved):

conftest.py
muddyfish: Yes

PR #98, opened 2023-12-13
Create CI release setup

IsaevIlya:
## Description
Create CI for building release wheels with github actions and cibuildwheel on EC2 instance.
In the current setup creation of release tag is used to initiate building of x86-x64 wheels; `run_cibuildwheel_on_ec2.sh` script is used for building arm wheels inside cibuildwheel container run inside EC2 instance.

--------


.github/workflows/ci.yml
```diff
@@ -2,7 +2,7 @@ name: Tests
 
 on:
   push:
-    branches: [ "main" ]
+    branches: [ "main", "tmp/distribution" ]
```
dnanuti, line 5: Is "tmp/distribution" a leftover here?

.github/workflows/ci.yml
```diff
@@ -16,4 +16,13 @@ jobs:
   python-checks:
     name: Python Checks
     uses: ./.github/workflows/python-checks.yml
+    permissions:
+      id-token: write
+      contents: read
 
+  wheels:
```
dnanuti, line 23: Do we want to run this on every PR/ push on main? Beyond adding the changes in main, what do you have in mind in terms of release workflow?

.github/workflows/generate_third_party_licenses.yml
```diff
@@ -0,0 +1,10 @@
+name: Generate THIRD-PARTY-LICENSES
+
+on:
+  push:
+    branches: [ "tmp/distribution" ]
```
dnanuti, line 5: This needs to be changed

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,96 @@
+name: Build Wheels
+
+on:
+  workflow_call:
+
+env:
+  CI_REGION: ${{ vars.S3_REGION }}
```
dnanuti, line 7: Not sure if we need all of these, if I'm not missing something, looks like we're using just `${{ vars.S3_REGION }}`

THIRD-PARTY-LICENSES
dnanuti, line 1: We do not want to add this file to the Github repository, please see the discussions here: https://github.com/awslabs/s3-connector-for-pytorch/pull/54

run_cibuildwheel_on_ec2.sh
```diff
@@ -0,0 +1,27 @@
...
+
+ROLE_NAME=$1
+REGION_NAME=$2
+BUCKET_NAME=$3
+PREFIX=$4
+EXPRESS_REGION_NAME=$5
+EXPRESS_BUCKET_NAME=$6
+
+FILE_NAME="tmp_cred.json"
+TOKEN=`curl -s -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600"` && curl -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/iam/security-credentials/${ROLE_NAME} >> ${FILE_NAME}
```
dnanuti, line 15: Can we please think of a nicer way to write this?

run_cibuildwheel_on_ec2.sh
```diff
@@ -0,0 +1,27 @@
...
+export AWS_SESSION_TOKEN=`cat ${FILE_NAME} | jq -r '.Token'`
+rm ${FILE_NAME}
+
+export CI_REGION=${REGION_NAME}
+export CI_BUCKET=${BUCKET_NAME}
+export CI_PREFIX=${PREFIX}
+export CI_EXPRESS_REGION=${EXPRESS_REGION_NAME}
+export CI_EXPRESS_BUCKET=${EXPRESS_BUCKET_NAME}
+
+cibuildwheel --output-dir wheelhouse --platform linux s3torchconnectorclient
```
dnanuti, line 27: nit: new line

s3torchconnectorclient/pyproject.toml
```diff
@@ -36,3 +37,50 @@ find = { where = ["python/src"] }
...
+[tool.setuptools]
+license-files = [ "LICENSE", "THIRD-PARTY-LICENSES", "NOTICE"]
+
+[tool.cibuildwheel]
+test-requires = ["./s3torchconnector[e2e]"]
+test-extras = "test"
+test-command = [
+    "pytest {package}/python/tst/",
+    "pytest {package}/../s3torchconnector/tst",
+    "CI_STORAGE_CLASS=EXPRESS_ONEZONE CI_REGION=${CI_EXPRESS_REGION} CI_BUCKET=${CI_EXPRESS_BUCKET}  pytest {package}/python/tst/integration",
```
dnanuti, line 50: These are running against S3 Express One Zone, where are we running against standard S3?
We also need CI_PREFIX to be set.

s3torchconnectorclient/pyproject.toml
```diff
@@ -36,3 +37,50 @@ find = { where = ["python/src"] }
...
+  "yum install -y dpkg",
+  "yum install -y fakeroot",
+  "yum install -y rpmdevtools",
+  "yum install -y tar",
+  "yum install -y wget",
+  "yum install -y devtoolset-10-gcc",
+  "yum install -y devtoolset-10-gcc-c++",
+  "yum install -y llvm-toolset-7.0-clang"
+]
+environment = { PATH="/opt/rh/llvm-toolset-7.0/root/usr/bin:/opt/rh/devtoolset-10/root/usr/bin:$HOME/.cargo/bin:$PATH", LD_LIBRARY_PATH="/opt/rh/llvm-toolset-7.0/root/usr/lib64:/opt/rh/devtoolset-10/root/usr/lib64:/opt/rh/devtoolset-10/root/usr/lib", CC="/opt/rh/devtoolset-10/root/usr/bin/gcc", CXX="/opt/rh/devtoolset-10/root/usr/bin/g++" }
```
dnanuti, line 86: This would need to be updated if we install different versions. Can we make it more generic?

dnanuti (dismissed):

.github/workflows/ci.yml
IsaevIlya: I am afraid we need that branch for building wheels, until we will change the flow to be relaying on release tags, but for that we need to spend more time.

.github/workflows/wheels.yml
IsaevIlya: We need to exposure all of them to be able to run integration tests for S3 Standart and S3 Express.

THIRD-PARTY-LICENSES
IsaevIlya: Thank you for catching it, I've missed that discussion before.

s3torchconnectorclient/pyproject.toml
IsaevIlya: The first two lines are running against S3 Standard.  CI_REGION/CI_BUCKET and their express counter part would be setup before executing that command. All of them will be exposed as part of `environment-pass` bellow in the file. Because we changed variable names for S3 Express integration tests to be the same as for S3 Standard, we need to override these variables, before starting tests for S3 Express, otherwise these variable will contains settings for S3 Standart.

s3torchconnectorclient/pyproject.toml
IsaevIlya: I'm not sure that it it is possible. It would need to be updated only in case mountpointclient change their build dependency.

run_cibuildwheel_on_ec2.sh
IsaevIlya: > Can we please think of a nicer way to write this?

Could you please explain the concept of `nicer` in that case? Is it to remove using of curl or splitting that command in few smaller steps?

.github/workflows/ci.yml
IsaevIlya: I think yes. We need to run integration tests before merging to main, but that step is part of building wheels. We need to build wheel to be able to run integration tests.

.github/workflows/ci.yml
IsaevIlya: Will replace that condition with `tags` condition.

.github/workflows/ci.yml
```diff
@@ -2,7 +2,7 @@ name: Tests
 
 on:
   push:
-    branches: [ "main" ]
+    tags: [ "v[0-9]+.[0-9]+.[0-9]+" ]
```
fuatbasik, line 5: This is a bit confusing. It looks like we will not be running part of the CI in push to main anymore. My understanding was, we like to keep everything as is but add a step `wheels` on push of tags.

.github/workflows/ci.yml
fuatbasik: We need to split them if thats the case. Integ-tests should run even when we are not building wheels. It should also run, with wheels, after we build them. But building wheels should not be a dependency of integration-testings. If we split these two concerns, we then probably want Integ-Tests to run on every PR and push to main. Wheels only build on Tags and Wheel build step is followed by a set if Integ-tests running with the artefacts of that step.

.github/workflows/generate_third_party_licenses_workflow.yml
```diff
@@ -0,0 +1,35 @@
+name: Generate THIRD-PARTY-LICENSES
```
fuatbasik, line 1: we are calling this workflow both on `push tags` as described in its own yml and `wheels` workflow. Do we need to call this twice?

run_cibuildwheel_on_ec2.sh
```diff
@@ -0,0 +1,32 @@
+if [ $# -ne 6 ]; then
```
fuatbasik, line 1: do we want this piece of code stay in the repo? I know it is operationally needed but does it really make sense to a collaborator? If yes, is this at the right format/structure currently?

s3torchconnectorclient/pyproject.toml
fuatbasik: Sorry bit confused, when running integration-tests here, are we running them using artefacts of the build_wheels step or running them from source-code. 

In other words what are we testing: 1/We built wheels from a commit that is passing integration-tests. 2/We built a wheel that passes all the tests-cases specified in our integration tests?

fuatbasik (commented):
Thanks a lot for the PR Ilya. I put a bunch of comments, most of them are actual questions.

.github/workflows/ci.yml
IsaevIlya: Canceled that changes and moved `tags` condition to `wheels.yml`

.github/workflows/generate_third_party_licenses_workflow.yml
IsaevIlya: We are using artifacts from this call for building arm wheels. When we will move building arm wheels to github actions, we can remove this call.

run_cibuildwheel_on_ec2.sh
IsaevIlya: This scrip could be used by collaborator to build wheels and run integration tests inside their EC2 instance. What we can improve here is to provide later some instruction on how to use it.

s3torchconnectorclient/pyproject.toml
IsaevIlya: Sorry for confusion, we have a separate workflow to run only integration tests in `python-integration.yml`. It is run for pushing to main and for creating PR to main, so I think we are fine with separate workflow that has no dependency on building wheels.

Here we are building wheels and testing that these wheels are passing all the tests.

THIRD-PARTY-LICENSES
dannycjones: Add that file to `.gitignore` (now or later)? Or put it in a directory with other build artifacts that'll be ignored?

.github/workflows/ci.yml
dnanuti: I think we should address that before merging this into main.

THIRD-PARTY-LICENSES
dnanuti: That branch had a stale version, that file is automatically generated when building wheels. I agree, we should add it to .gitignore to prevent such cases. @IsaevIlya would you please update that?

run_cibuildwheel_on_ec2.sh
dnanuti: I am not particularly fond of using the IP address

s3torchconnectorclient/pyproject.toml
dnanuti: > test-command = [
    "pytest {package}/python/tst/",
    "pytest {package}/../s3torchconnector/tst",
    "CI_STORAGE_CLASS=EXPRESS_ONEZONE CI_REGION=${CI_EXPRESS_REGION} CI_BUCKET=${CI_EXPRESS_BUCKET}  pytest {package}/python/tst/integration",
    "pytest {package}/../s3torchconnector/tst/e2e",
]
What I was trying to call out is that the way this is currently achieved is a bit confusing, although correct, as we are selecting just the subset of tests dependent on S3 Express setup. Do you think it would be worth to have a more verbose approach here?
i.e. unit tests command appears once, integration / e2e appears command appears once for S3, once for S3 Express.

run_cibuildwheel_on_ec2.sh
IsaevIlya: > I am not particularly fond of using the IP address

That ip address is official way to access the EC2 instance metadata from within instance itself: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html

.github/workflows/generate_third_party_licenses.yml
```diff
@@ -0,0 +1,10 @@
+name: Generate THIRD-PARTY-LICENSES
+
```
dnanuti, line 2: Sorry for missing this previously, but isn't this working if we simply add this in `github/workflows/generate_third_party_licenses_workflow.yml` and remove this file? 

```
push:
    tags: [ "v[0-9]+.[0-9]+.[0-9]+" ]
```

IsaevIlya: We are generating 3rd party license file in two place at the moment. One action is the build of wheels for  x86, the second action create only license file. The second action is needed that we could start build of arm wheels before all the x86 wheels would be build.

dnanuti: I understood that, what I meant is does it work to have something like in the next file and remove this one? 
```
on:
  push:
    tags: [ "v[0-9]+.[0-9]+.[0-9]+" ]
  workflow_call:
```

dnanuti (approved):

Issue #102, opened 2024-01-09
Error when importing laion-art dataset


awsankur:
### s3torchconnector version

1.1.0

### s3torchconnectorclient version

1.1.0

### AWS Region

us-west-2

### Describe the running environment

Running in EC2 m5.8xlarge Amazon Linux 2

### What happened?

I am following the example notebook here: [https://github.com/awslabs/s3-connector-for-pytorch/blob/main/examples/Getting%20s[…]ed%20with%20the%20Amazon%20S3%20Connector%20for%20PyTorch.ipynb](https://github.com/awslabs/s3-connector-for-pytorch/blob/main/examples/Getting%20started%20with%20the%20Amazon%20S3%20Connector%20for%20PyTorch.ipynb)

I need to modify it to load a dataset where multiple images are sharded in hundreds of .tar files. An example dataset is the laion-art dataset.
I am using the following code:

```
import s3torchconnector

from PIL import Image
import webdataset
import itertools
import io
import torch
import torchvision


def shard_to_dict(object):
    return {"url": object.key, "stream": object}

IMAGES_URI = "s3://laion-art/laion-art-data/tarfiles_reorganized/task0000/"

REGION = "us-west-2"

s3_dataset = s3torchconnector.S3MapDataset.from_prefix(IMAGES_URI, region=REGION, transform=shard_to_dict)

dataset = webdataset.tariterators.tar_file_expander(s3_dataset)

loader = torch.utils.data.DataLoader(dataset, batch_size=4)
for batch in loader:
    print(batch)
```

But I get the error:
```TypeError: object of type 'generator' has no len()```

### Relevant log output

```shell
(pt-nightlies) ubuntu@ip-10-0-55-158:/apps$ python3 s3connector.py
Traceback (most recent call last):
  File "/apps/s3connector.py", line 23, in <module>
    for batch in loader:
  File "/apps/.conda/envs/pt-nightlies/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 631, in __next__
    data = self._next_data()
  File "/apps/.conda/envs/pt-nightlies/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 674, in _next_data
    index = self._next_index()  # may raise StopIteration
  File "/apps/.conda/envs/pt-nightlies/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 621, in _next_index
    return next(self._sampler_iter)  # may raise StopIteration
[...]

PR #103, opened 2024-01-10
Be more explicit on how to use tar datasets with DataLoader in our example

muddyfish:
## Description

In https://github.com/awslabs/s3-connector-for-pytorch/issues/102, we found it wasn't obvious how to integrate `tar_file_expander` with `DataLoader`, and our example for tar file extraction skipped over it. 
This PR adds an explicit example for showing how to wrap an iterable such as one returned by `tar_file_expander` into a dataloader. 

## Additional context

No changes to the codebase.

## Related items

https://github.com/awslabs/s3-connector-for-pytorch/issues/102

## Testing

- Executed code for new changes and verified they behaved as expected.
- Viewed modified Jupyter notebook and saw that it was rendered correctly.

--------

### DIFF
--- a/examples/Getting started with the Amazon S3 Connector for PyTorch.ipynb
+++ b/examples/Getting started with the Amazon S3 Connector for PyTorch.ipynb
@@ -643,7 +643,7 @@
     {
      "data": {
       "text/plain": [
-       "'2022.001.0000.00.DISC.01.jpg.image.jpg'"
+       "['2022.001.0000.00.DISC.01.jpg.image.jpg']"
       ]
      },
      "metadata": {},
@@ -663,7 +663,7 @@
     {
      "data": {
       "text/plain": [
-       "'2022.001.0320.00.DISC.01.jpg.image.jpg'"
+       "['2022.001.0320.00.DISC.01.jpg.image.jpg']"
       ]
      },
      "metadata": {},
@@ -683,7 +683,7 @@
     {
      "data": {
       "text/plain": [
-       "'2022.001.0640.00.DISC.01.jpg.image.jpg'"
+       "['2022.001.0640.00.DISC.01.jpg.image.jpg']"
       ]
      },
      "metadata": {},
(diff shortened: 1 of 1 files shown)


IsaevIlya (approved):

examples/Getting started with the Amazon S3 Connector for PyTorch.ipynb
```diff
@@ -643,7 +643,7 @@
     {
      "data": {
       "text/plain": [
-       "'2022.001.0000.00.DISC.01.jpg.image.jpg'"
+       "['2022.001.0000.00.DISC.01.jpg.image.jpg']"
```
dnanuti, line 646: I am probably missing something, but why do we need this change?

examples/Getting started with the Amazon S3 Connector for PyTorch.ipynb
```diff
@@ -745,12 +745,15 @@
...
+    "tar_dataset = webdataset.tariterators.tar_file_expander(s3_dataset)\n",
+    "dataset = torchdata.datapipes.iter.IterableWrapper(tar_dataset, deepcopy=False)\n",
     "\n",
-    "for sample in itertools.islice(dataset, 0, 100, 20):\n",
+    "loader = torch.utils.data.DataLoader(dataset)\n",
+    "\n",
+    "for sample in itertools.islice(loader, 0, 100, 20):\n",
     "    display(sample[\"fname\"])\n",
-    "    display(Image.open(io.BytesIO(sample[\"data\"])).reduce(8))"
+    "    display(Image.open(io.BytesIO(sample[\"data\"][0])).reduce(8))"
```
dnanuti, line 756: This should not be needed if we decide we don't need the change above.

dnanuti (changes requested):

examples/Getting started with the Amazon S3 Connector for PyTorch.ipynb
muddyfish: Using `DataLoader` instead of a plain iterator means we're using PyTorch's batch system. 
This means that our data will always be batched, so the behaviour on iteration changes. The example here uses size 1, so to get the data we want out, you have to do indexing.

dnanuti (approved):

PR #104, opened 2024-01-11
Add public version attributes

dnanuti:
Make version public for s3torchconnector and s3torchconnectorclient

## Description

Add `__version__` to the public interface for both s3torchconnector and s3torchconnectorclient, to be easier for us to verify which version a customer is using.

## Testing

- Unit tests
--------

### DIFF
--- a/s3torchconnector/src/s3torchconnector/__init__.py
+++ b/s3torchconnector/src/s3torchconnector/__init__.py
@@ -10,6 +10,7 @@ from .s3writer import S3Writer
 from .s3iterable_dataset import S3IterableDataset
 from .s3map_dataset import S3MapDataset
 from .s3checkpoint import S3Checkpoint
+from ._version import __version__
 
 __all__ = [
     "S3IterableDataset",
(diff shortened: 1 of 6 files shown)


s3torchconnector/src/s3torchconnector/__init__.py
```diff
@@ -18,4 +19,5 @@
     "S3Reader",
     "S3Writer",
     "S3Exception",
+    "__version__",
```
IsaevIlya, line 22: Would it make sense to export version with unique name to prevent possible confusions, because of names clashing?

dnanuti: Not really, `__version__` is the PEP8 standard for naming this: https://peps.python.org/pep-0008/#module-level-dunder-names

If we need to import them both as part of a file we can have import aliases:
```
from s3torchconnector import __version__ as connector_version
from s3torchconnectorclient import  __version__ as client_version
```

IsaevIlya (approved):

PR #106, opened 2024-01-12
Add Slack Notifier workflow

muddyfish:
## Description

Individual notifier workflows for:
- PR creation
- New issues
- New comments on issues

Each notifier workflow has an associated slack webhook url secret added to the repository.

## Additional context

No code changes

## Testing

Created a test PR, issue, and issue comments and verified that Slack messages are sent.

--------

### DIFF
--- a/.github/workflows/notify-issue-comment.yml
+++ b/.github/workflows/notify-issue-comment.yml
@@ -1,0 +1,20 @@
+name: Issue Comment Slack Notifier
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  notify:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send notification to Slack
+        uses: slackapi/slack-github-action@v1.24.0
+        with:
+          payload: |
+            {
+              "comment_url": "${{ github.event.comment.html_url }}"
+            }
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL_ISSUE_COMMENT }}
+          SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
--- a/.github/workflows/notify-issue.yml
+++ b/.github/workflows/notify-issue.yml
@@ -1,0 +1,20 @@
+name: Issue Slack Notifier
+
+on:
+  issues:
+    types: [opened]
+
+jobs:
+  notify:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send notification to Slack
+        uses: slackapi/slack-github-action@v1.24.0
+        with:
+          payload: |
+            {
+              "issue_url": "${{ github.event.issue.html_url }}"
+            }
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL_ISSUE }}
+          SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
--- a/.github/workflows/notify-issue.yml
+++ b/.github/workflows/notify-issue.yml
@@ -2,7 +2,7 @@ name: Issue Slack Notifier
 
 on:
   issues:
-    types: [opened]
+    types: [opened, reopened, edited]
 
 jobs:
   notify:
@@ -13,6 +13,7 @@ jobs:
         with:
           payload: |
             {
+              "action": "${{ github.event.action }}",
               "issue_url": "${{ github.event.issue.html_url }}"
             }
         env:
--- a/.github/workflows/notify-pr.yml
+++ b/.github/workflows/notify-pr.yml
@@ -1,0 +1,21 @@
+name: Pull Request Slack Notifier
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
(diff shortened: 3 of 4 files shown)


.github/workflows/notify-pr.yml
```diff
@@ -0,0 +1,21 @@
...
+  notify:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send notification to Slack
+        uses: slackapi/slack-github-action@v1.24.0
+        with:
+          payload: |
+            {
+              "action": "${{ github.event.action }}",
+              "url": "${{ github.event.pull_request.html_url || github.event.head_commit.url }}"
```
IsaevIlya, line 17: Should we include title and status? Status to inform on the closed PR too.

.github/workflows/notify-issue.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  notify:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send notification to Slack
+        uses: slackapi/slack-github-action@v1.24.0
+        with:
+          payload: |
+            {
+              "issue_url": "${{ github.event.issue.html_url }}"
```
IsaevIlya, line 16: Should we include title?

.github/workflows/notify-issue-comment.yml
```diff
@@ -0,0 +1,20 @@
+name: Issue Comment Slack Notifier
+
+on:
+  issue_comment:
```
IsaevIlya, line 4: Based on the doc:
`The issue_comment event occurs for comments on both issues and pull requests.`
If we want that workflow to trigger only for the comment on issue, we need to add additional check:
`if: ${{ !github.event.issue.pull_request }}`

But I'm thinking that comments on PR could be beneficial too. We could have a separate statement for comments on PR with the comment itself.

.github/workflows/notify-issue-comment.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  notify:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send notification to Slack
+        uses: slackapi/slack-github-action@v1.24.0
+        with:
+          payload: |
+            {
+              "comment_url": "${{ github.event.comment.html_url }}"
```
IsaevIlya, line 16: Should we add the title of issue for which comment was added?

.github/workflows/notify-pr.yml
```diff
@@ -0,0 +1,21 @@
+name: Pull Request Slack Notifier
+
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize]
```
IsaevIlya, line 5: Should we add `closed` too to inform when PR is merged?

IsaevIlya (changes requested):

.github/workflows/notify-issue.yml
muddyfish: My thoughts were that it wasn't needed as Slack automatically embeds the issue title and initial content as part of it's link unfurling

.github/workflows/notify-issue-comment.yml
muddyfish: I think it would be good to have Slack messages on both. Do you think any changes are needed here?

.github/workflows/notify-issue-comment.yml
muddyfish: Again, the issue gets automatically unfurled and shows the start of the original issue. This shouldn't be needed

.github/workflows/notify-issue.yml
```diff
@@ -0,0 +1,20 @@
+name: Issue Slack Notifier
+
+on:
+  issues:
+    types: [opened]
```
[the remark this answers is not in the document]
muddyfish: I have no strong opinion here

.github/workflows/notify-issue.yml
IsaevIlya: Indeed. I haven't though about that. It could be an issue with private PR, but we won't have such one.
Let's try without a title and have a look how slack will manage preview.

.github/workflows/notify-pr.yml
muddyfish: The action parameter includes `opened, reopened, synchronize`, and the title is included in the unfurling.

.github/workflows/notify-issue-comment.yml
IsaevIlya: If we are going to use both, then small nitpick: let's change the name of workflow to reflect that.

Separately, I've tried to post link on the comments for an issue and for a PR. In case of PR's comment slack is not trying to build preview. Let's add separate statement for PR's comments to include the comment text.

dnanuti: We should be notified for both pull request comments and issue comments. I think we are a bit confused as we haven't seen examples of how these notifications would look like. Is it possible to have a couple of test notifications on the team's channel?

dnanuti (dismissed):

IsaevIlya (approved):

PR #108, opened 2024-01-18
Expose function to enable debug logging

dnanuti:
Draft function to enable CRT logs for debug purposes

## Description

We need a user friendly way to enable logs for debugging purposes. This is a rough method to do this. 

## Testing
- Ran an E2E test locally

--------

### DIFF
--- a/s3torchconnector/src/s3torchconnector/__init__.py
+++ b/s3torchconnector/src/s3torchconnector/__init__.py
@@ -1,6 +1,6 @@
 #  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #  // SPDX-License-Identifier: BSD
-from s3torchconnectorclient import S3Exception
+from s3torchconnectorclient import S3Exception, enable_debug_logging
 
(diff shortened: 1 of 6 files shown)


s3torchconnector/src/s3torchconnector/__init__.py
```diff
@@ -1,6 +1,6 @@
 #  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #  // SPDX-License-Identifier: BSD
-from s3torchconnectorclient import S3Exception
+from s3torchconnectorclient import S3Exception, enable_debug_logging
```
dnanuti, line 3: Thinking out loud - perhaps we also need a wrapper on the Python side of things that takes in a logging config and a log level and configures the logging module like below, but with the passed in parameters:
```
logging.basicConfig(
    format="%(levelname)s %(name)s %(asctime)-15s %(filename)s:%(lineno)d %(message)s"
)
logging.getLogger().setLevel(LOG_TRACE)
```

PR #125, opened 2024-02-01
Fix wheel generation artefact name

muddyfish:
## Description

Stop using a hardcoded artefact name, and use the same generation process ORT uses for it's own artefact name generation.

## Additional context

Third party license generation/wheel workflows were not working in forked repositories. This change fixes this.

## Testing

Tested on a workflow in a private fork

--------

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -35,6 +35,16 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Initialise environment
+        shell: bash
+        env:
+          REPO_NAME: ${{ github.event.repository.name }}
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/wheels.yml
```diff
@@ -48,7 +58,7 @@ jobs:
       # https://github.com/actions/upload-artifact/issues/478
       - uses: actions/download-artifact@v3
         with:
-          name: "ort-results-s3connectorforpytorch--"
+          name: ${{ ARTIFACT_NAME }}
```
IsaevIlya, line 61: Why can't we use hardcoded name here? For what reason we want general solution here?

muddyfish: In the private repository, we were wanting to get all of the workflows working, including the wheel generation workflow.

muddyfish: ORT uses the repository name to determine what the uploaded artefact will be called, so hardcoding it doesn't work if we want it working between multiple repositories

IsaevIlya (approved):

PR #128, opened 2024-02-06
Add additional logging for construction and HeadObject requests

muddyfish:
## Description

Adds logging at INFO level whenever a Dataset is constructed
Adds logging at DEBUG level whenever a HeadObject request is explicitly made from the s3torchconnector side. 

## Additional context

No breaking changes

## Testing

Added unit tests, ran a small example script and saw this in the output:
```
INFO s3torchconnector.s3map_dataset 2024-02-06 16:06:04,593 s3map_dataset.py:98 - Building S3MapDataset from_prefix
```

--------

### DIFF
--- a/s3torchconnector/src/s3torchconnector/s3iterable_dataset.py
+++ b/s3torchconnector/src/s3torchconnector/s3iterable_dataset.py
@@ -2,6 +2,7 @@
 #  // SPDX-License-Identifier: BSD
 from functools import partial
 from typing import Iterator, Any, Union, Iterable, Callable
+import logging
 
 import torch.utils.data
 
@@ -13,6 +14,8 @@ from ._s3dataset_common import (
     get_objects_from_uris,
     get_objects_from_prefix,
 )
+
+log = logging.getLogger(__name__)
 
 
 class S3IterableDataset(torch.utils.data.IterableDataset):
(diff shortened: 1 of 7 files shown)


IsaevIlya (changes requested):
Thank you for improving our log visibility.
I'm wondering, we are providing visibility on to `head_object` calls, why not to do it for `get_object`, `put_object` and `list_objects` requests at the same time?

muddyfish:
> Thank you for improving our log visibility. I'm wondering, we are providing visibility on to `head_object` calls, why not to do it for `get_object`, `put_object` and `list_objects` requests at the same time?

I can do it for `list_objects`, but `get_object` and `put_object` don't match one to one with calls to S3. If you think it's still useful, we can still add them with that knowledge

IsaevIlya:
> I can do it for `list_objects`, but `get_object` and `put_object` don't match one to one with calls to S3. If you think it's still useful, we can still add them with that knowledge
I totally agree on your comment, but I think it would be nice to know when we are explicitly calling these methods, even thought we can't translate them 1-to-1 to requests to S3.

s3torchconnector/src/s3torchconnector/s3iterable_dataset.py
```diff
@@ -83,6 +87,7 @@ def from_prefix(
         Raises:
             S3Exception: An error occurred accessing S3.
         """
+        log.info(f"Building {cls.__name__} from_prefix")
```
dnanuti, line 90: Should we also include the prefix here? 
For objects_uris is too much to log, but prefix might be useful.

dnanuti (dismissed):

dnanuti:
nit: it's failing Python lints

IsaevIlya (dismissed):

IsaevIlya (approved):

dnanuti (approved):

PR #129, opened 2024-02-07
Add delete_object support in s3torchconnectorclient

muddyfish:
## Description

Adds `delete_object` support to `s3torchconnectorclient`, with associated unit and integration tests. 

## Additional context

No breaking changes.

## Testing

Ran unit and integration tests.

Created additional unit and integration tests for delete support

--------


dnanuti:
Please provide more details in the description! Thank you 😄

s3torchconnectorclient/python/tst/unit/test_mountpoint_s3_client.py
```diff
@@ -288,6 +288,40 @@ def test_mountpoint_client_pickles():
...
+
+
+def test_delete_object_already_deleted():
+    mock_client = MockMountpointS3Client(REGION, MOCK_BUCKET)
+    client = mock_client.create_mocked_client()
+
+    client.delete_object(MOCK_BUCKET, "hello_world.txt")
+
+
+def test_delete_object_non_existant_bucket():
```
dnanuti, line 316: nit: typo, existent*

s3torchconnectorclient/python/tst/unit/test_mountpoint_s3_client.py
```diff
@@ -288,6 +288,40 @@ def test_mountpoint_client_pickles():
     assert client.no_sign_request == loaded.no_sign_request == expected_no_sign_request
 
 
+@pytest.mark.parametrize(
+    "key",
+    [
+        "hello_world.txt",
```
dnanuti, line 294: Do we want to add more parameters here? If not, we can drop parametrized.

s3torchconnectorclient/python/tst/integration/test_mountpoint_s3_integration.py
```diff
@@ -258,6 +258,40 @@ def test_head_object(sample_directory):
...
+            client.get_object(
+                sample_directory.bucket, f"{sample_directory.prefix}hello_world.txt"
+            )
+        )
+
+    assert str(e.value) == "Service error: The key does not exist"
+
+
+def test_delete_object_does_not_exist(empty_directory):
+    client = MountpointS3Client(empty_directory.region, TEST_USER_AGENT_PREFIX)
```
IsaevIlya, line 280: Could we make our intention for this unit test more clear, please? I suppose our expectation here, that call won't throw exception.

s3torchconnectorclient/python/tst/integration/test_mountpoint_s3_integration.py
```diff
@@ -258,6 +258,40 @@ def test_head_object(sample_directory):
         assert object_info.etag == expected_etag
 
 
+@pytest.mark.parametrize("delete_calls", [1, 2])
+def test_delete_object(delete_calls: int, sample_directory):
+    client = MountpointS3Client(sample_directory.region, TEST_USER_AGENT_PREFIX)
+    for i in range(delete_calls):
```
IsaevIlya, line 264: Could you please explain why do we need to call delete_object twice here?

muddyfish: The idea was to verify that `delete_object` does not throw in case the object is already deleted, but it's also tested in `test_delete_object_does_not_exist`, so I'm happy to remove it here

s3torchconnectorclient/python/tst/integration/test_mountpoint_s3_integration.py
muddyfish: I'll add a comment on it, but we're wanting to assert that no exception is raised which is done implicitly by our testing framework.

s3torchconnectorclient/python/tst/unit/test_mountpoint_s3_client.py
```diff
@@ -288,6 +288,40 @@ def test_mountpoint_client_pickles():
...
+    ],
+)
+def test_delete_object(key: str):
+    mock_client = MockMountpointS3Client(REGION, MOCK_BUCKET)
+    mock_client.add_object(key, b"data")
+    client = mock_client.create_mocked_client()
+
+    client.delete_object(MOCK_BUCKET, key)
+
+    with pytest.raises(S3Exception) as e:
```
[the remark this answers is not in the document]
muddyfish: Is this better?
```
with pytest.raises(S3Exception, match="Service error: The key does not exist")
```

IsaevIlya (approved):

dnanuti (approved):

PR #132, opened 2024-02-08
Lightning support

muddyfish:
## Description

- Adds support for PyTorch Lightning style checkpoints implementing `CheckpointIO`, with implementations for `save_checkpoint`, `load_checkpoint`, and `remove_checkpoint`
- Adds optional dependency on `lightning`
- The implementation only supports `str` as a path type.

## Additional context

No breaking changes

## Testing

- Adds unit tests for `S3LightningCheckpoing`
- Adds tests that validates the error we get when importing `s3torchconnector.lightning` if `lightning` isn't installed
- Ensures existing unit tests don't fail in case lightning isn't installed
- Ran code coverage on `S3LightningCheckpoing` and verified 100% coverage within `s3_lightning_checkpoint.py`

--------


s3torchconnector/tst/unit/lightning/test_s3_lightning_checkpoint.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+@given(python_primitives, byteorders)
+def test_lightning_checkpointing_saves_python_primitives(data, byteorder):
+    _test_save(data, byteorder)
+
+
+@given(byteorders)
+def test_lightning_checkpointing_saves_tensor(byteorder):
+    tensor = torch.tensor([[0.1, 1.2], [2.2, 3.1], [4.9, 5.2]])
```
dnanuti, line 36: Can we please use `torch.rand(...)` here and for the other occurences? 
https://pytorch.org/docs/stable/generated/torch.rand.html

s3torchconnector/tst/unit/_checkpoint_utils.py
dnanuti, line 1: Can we think of a more specific name for this?

s3torchconnector/tst/unit/lightning/test_s3_lightning_checkpoint.py
```diff
@@ -0,0 +1,146 @@
...
+    with pytest.raises(S3Exception) as error:
+        client.get_object(TEST_BUCKET, TEST_KEY).read()
+    assert str(error.value) == "Service error: The key does not exist"
+
+
+@pytest.mark.parametrize(
+    "checkpoint_method_name, kwargs",
+    [
+        ("save_checkpoint", {"path": Path(), "checkpoint": None}),
+        ("load_checkpoint", {"path": Path()}),
```
dnanuti, line 88: Would probably be worth to pass some strings in the Path initialization too for completeness, even though the validation now is just a type check.

s3torchconnector/src/s3torchconnector/lightning/s3_lightning_checkpoint.py
```diff
@@ -0,0 +1,82 @@
...
+        bucket, key = parse_s3_uri(path)
+        with self._client.put_object(bucket, key) as s3writer:
+            torch.save(checkpoint, s3writer)
+
+    def load_checkpoint(
+        self,
+        path: str,
+        map_location: Optional[Any] = None,
+    ) -> Dict[str, Any]:
+        """Load checkpoint from a path when resuming or loading ckpt for test/validate/predict stages.
```
dnanuti, line 45: `from an S3 location` perhaps?
I would avoid path since we are supporting valid S3 URIs in string format.

.github/workflows/python-checks.yml
```diff
@@ -57,7 +57,13 @@ jobs:
       - name: s3torchconnectorclient unit tests
         run: pytest s3torchconnectorclient/python/tst/unit --hypothesis-profile ci --hypothesis-show-statistics
       - name: s3torchconnector unit tests
-        run: pytest s3torchconnector/tst/unit --hypothesis-profile ci --hypothesis-show-statistics
+        run: pytest s3torchconnector/tst/unit --ignore s3torchconnector/tst/unit/lightning --hypothesis-profile ci --hypothesis-show-statistics
+
+      - name: Install Lightning dependency
+        run: |
+          python -m pip install -e "s3torchconnector[lightning]"
```
dnanuti, line 64: Latest Lightning version is 2.1.4. Should we try to have a matrix here with lightning versions that we specifically install and run tests on?

s3torchconnector/tst/unit/lightning/test_s3_lightning_checkpoint.py
```diff
@@ -0,0 +1,146 @@
...
+        checkpoint_method(**kwargs)
+
+
+def test_teardown():
+    s3_lightning_checkpoint = S3LightningCheckpoint(TEST_REGION)
+    s3_lightning_checkpoint.teardown()
+    # Assert no exception is thrown - implicit
+
+
+def _test_save(
```
IsaevIlya, line 105: Could we add an unit test for `save_checkpoint` when `S3LightningCheckpoint` is wrapped by [AsyncCheckpointIO](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.plugins.io.AsyncCheckpointIO.html)?

.github/workflows/python-checks.yml
muddyfish: I'm not sure, though I think if we do decide to do this, we should do it in a separate PR, possibly along with a matrix of pytorch versions

s3torchconnector/tst/unit/lightning/test_s3_lightning_checkpoint.py
```diff
@@ -6,62 +6,105 @@
...
+    tensor = torch.rand(2, 4)
+    _test_save(client, lightning_checkpoint, tensor, byteorder, equal=torch.equal)
 
 
+@hypothesis.settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
 @given(byteorders)
-def test_lightning_checkpointing_saves_tensor(byteorder):
-    tensor = torch.tensor([[0.1, 1.2], [2.2, 3.1], [4.9, 5.2]])
-    _test_save(tensor, byteorder, equal=torch.equal)
+def test_async_lightning_checkpointing_saves_tensor(client, lightning_checkpoint, byteorder):
```
IsaevIlya, line 64: Should there be `async_lightning_checkpoint` as input parameter?

s3torchconnector/tst/unit/lightning/test_s3_lightning_checkpoint.py
```diff
@@ -6,62 +6,105 @@
...
+    tensor = torch.rand(2, 4)
+    _test_load(client, lightning_checkpoint, tensor, byteorder, equal=torch.equal)
 
 
+@hypothesis.settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
 @given(byteorders)
-def test_lightning_checkpointing_loads_tensor(byteorder):
-    tensor = torch.tensor([[0.1, 1.2], [2.2, 3.1], [4.9, 5.2]])
-    _test_load(tensor, byteorder, equal=torch.equal)
+def test_lightning_checkpointing_async_loads_tensor(client, async_lightning_checkpoint, byteorder):
```
IsaevIlya, line 97: nit: the name is inconsistent with `test_async_lightning_checkpointing_saves_tensor`
We may not need that test at all, as `AsyncCheckpointIO` doesn't provide async method for loading path.

IsaevIlya (changes requested):

s3torchconnector/src/s3torchconnector/lightning/s3_lightning_checkpoint.py
```diff
@@ -0,0 +1,82 @@
...
+        self._client.delete_object(bucket, key)
+
+    def teardown(self) -> None:
+        """This method is called to teardown the process."""
+        pass
+
+    @staticmethod
+    def _validate_path(path: str) -> None:
+        if not isinstance(path, str):
+            raise TypeError(f"argument must be a string, not {type(path).__name__!r}")
```
dnanuti, line 82: We might want to make this more verbose - i.e. "Path is not supported. Argument must be a valid s3_uri string: s3://bucket/prefix".

muddyfish: Changing this to 
```
"{type(path).__name__!r} is not a supported type for 'path'. Must be a string formatted as an S3 uri."
```

IsaevIlya (approved):

s3torchconnector/src/s3torchconnector/lightning/s3_lightning_checkpoint.py
```diff
@@ -0,0 +1,84 @@
...
+    def save_checkpoint(
+        self,
+        checkpoint: Dict[str, Any],
+        path: str,
+        storage_options: Optional[Any] = None,
+    ) -> None:
+        """Save model/training states as a checkpoint file through state-dump and upload to S3.
+
+        Args:
+            checkpoint: dict containing model and trainer state
```
dnanuti, line 30: nit: consistency

> checkpoint: dict
Should be checkpoint: Dict ... 
Same for path description.

dnanuti (dismissed):

dnanuti:
We should merge this only after running the tests. Thanks! 👍

PR #134, opened 2024-02-09
Run tests on pull requests to feature branches

muddyfish:
## Description

Run CI workflows on pull requests to feature branches, as we're working on merging lightning support to a feature branch, but tests aren't being ran on it.

## Additional context

No breaking changes

## Testing

--------

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: Tests
 
 on:
   push:
-    branches: [ "main" ]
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
```
dnanuti, line 4: We want the same behaviour as main, so we probably want on push as well.

dnanuti (approved):

PR #135, opened 2024-02-09
Provide tooling for running benchmarks against pytorch connectors

vishalg0wda:
## Description
2 CLIs are provided:
 - `examples/benchmark/benchmark.py`: used to run benchmarks
 - `examples/benchmark/datagen.py`: used to synthesize datasets against which benchmarks can run

Some reference benchmarking scenarios have been defined that illustrate how similar experiments can be
defined. `examples/benchmark/README.md` has further instructions.

## Testing
- Validated both scripts are functional by running them against a fresh EC2 instance.

--------


examples/benchmark/models.py
```diff
@@ -0,0 +1,194 @@
...
+        """Train the model using given dataloader for number of epochs"""
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def save(self):
+        """Save checkpoint"""
+        raise NotImplementedError
+
+
+# This is not really a training model as it does not train anything
```
fuatbasik, line 39: nit: We probably should make these comments python-style i.e. Below the class description, with a ''' block.

examples/benchmark/README.md
```diff
@@ -0,0 +1,163 @@
...
+the end to end training performance. To achieve this, we split configuration to 4 pieces,
+namely; `dataset`, `dataloader`, `checkpoint`, `training`.
+
+The `dataset` configuration keeps information about where data resides. While we support sharded objects, only loading
+from TAR objects supported currently. See [a](#benchmarking-s3-connector-for-pytorch)
+The `dataloader` configuration contains dataloading specific setup. `kind` specify which PyTorch dataset to
+use (`s3iterabledataset`, `s3mapdataset`, `fsspec`) while
+`batch_size`, `num_workers` specify
+relevant [PyTorch DataLoader](https://pytorch.org/tutorials/beginner/basics/data_tutorial.html) parameters. `training`
+configuration specify what model to learn and how many epochs to execute the training for. Currently, we support two
```
fuatbasik, line 21: nit: support -> implemented

examples/benchmark/README.md
```diff
@@ -0,0 +1,163 @@
...
+a [g5.2xlarge](https://aws.amazon.com/ec2/instance-types/g5/)),
+choosing
+the [AWS Deep Learning AMI GPU PyTorch 2.0.1 (Amazon Linux 2)](https://aws.amazon.com/releasenotes/aws-deep-learning-ami-gpu-pytorch-2-0-amazon-linux-2/)
+as your AMI.
+
+You will need an S3 Bucket with training data and update the `dataset` configuration with correct prefix and region.
+
+Then from this directory, install the dependencies for the example code:
+
+    python -m pip install -r requirements.txt
```
fuatbasik, line 68: nit: Should we add virtual environment creation comments too?

examples/benchmark/README.md
```diff
@@ -0,0 +1,163 @@
...
+                           s3://<BUCKET>/1k_256x256_16Mib_sharded/
+  --region TEXT            Region where the S3 bucket is hosted.  [default:
+                           us-east-1]
+  --help                   Show this message and exit.
+
+```
+
+Here are some sample dataset configurations that we ran our benchmarks against:
+
+- `n: 20k, resolution: 496x387`
```
fuatbasik, line 124: is n expecting a string or a number? How does this parameter look to generate 1000000 images, 1M?

fuatbasik (commented):
Thanks a lot for this PR. I put some comments but none of them are blocking. For the future comments, it would be great if you could organise the commit messages following these 7 rules: https://cbea.ms/git-commit/

examples/benchmark/README.md
```diff
@@ -0,0 +1,163 @@
...
+The goal of this component is to create an easy to re-produce and extend performance benchmarking. This way, user can
+experiment with different settings and decide to best parameters for their workloads, before committing to a setup.
+
+By managing complex configuration space with [Hydra](https://hydra.cc/) this module splits configuration of multiple
+components of a training pipeline into smaller pieces to create custom
+training pipelines. This way, one can mix and match these configurations and observe the impact of different components
+to
+the end to end training performance. To achieve this, we split configuration to 4 pieces,
+namely; `dataset`, `dataloader`, `checkpoint`, `training`.
+
```
dnanuti, line 14: nit: use subsections here for `dataset`, `dataloader`, `checkpoint`, `training`

examples/benchmark/README.md
```diff
@@ -0,0 +1,163 @@
+# Benchmarking S3 Connector for PyTorch
```
dnanuti, line 1: There a bunch of things that can be improved from a grammar point of view. Can we please run a grammar check on this?

examples/benchmark/README.md
dnanuti: More generic: what are the possible value sets for each option?

examples/benchmark/benchmark.py
```diff
@@ -0,0 +1,113 @@
...
+from benchmark_utils import ResourceMonitor
+from models import Entitlement, ViT
+
+
+@hydra.main(version_base=None, config_path="configuration")
+def run_experiment(config: DictConfig):
+    if config.training.model == "entitlement":
+        model = Entitlement()
+    elif config.training.model == "vit":
+        model = ViT(1000, config.checkpoint)
```
dnanuti, line 19: Why are we hardcoding 1000?

examples/benchmark/models.py
```diff
@@ -0,0 +1,194 @@
...
+# Instead, this model loads data from S3, converts it to an image and
+# do nothing with it.
+class Entitlement(ModelInterface):
+    def __init__(self, num_labels: int = None):
+        self.name = "Entitlement"
```
muddyfish, line 44: Name should be in the base class if it's an assumed attribute

examples/benchmark/benchmark.py
```diff
@@ -0,0 +1,113 @@
...
+    else:
+        raise Exception(f"unknown model {config.training.model}")
```
muddyfish, line 21: I'd move the choosing the model bit to a function

examples/benchmark/benchmark.py
```diff
@@ -0,0 +1,113 @@
...
+
+    dataset = make_dataset(
+        config.dataloader.kind,
+        config.dataset.sharding,
+        config.dataset.prefix_uri,
+        config.dataset.region,
+        model.load_sample,
+        num_workers=config.dataloader.num_workers,
```
muddyfish, line 29: Why is num_workers the only keyword argument here?

examples/benchmark/benchmark_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    ):
+        self.training_time = training_time
+        self.num_samples = num_samples
+        self.throughput = self.num_samples / self.training_time
+        self.resource_data = {}
+        self.avg_resource_data = {}
+        self.checkpoint_results = checkpoint_results
+
+
+class ResourceMonitor:
```
muddyfish, line 30: Please define `__enter__` and `__exit__` so this can be used in `with` clauses

examples/benchmark/benchmark.py
```diff
@@ -0,0 +1,113 @@
...
+
+def make_dataset(
+    kind: str,
+    sharded: bool,
+    prefix_uri: str,
+    region: str,
+    load_sample,
+    num_workers: int,
+):
+    if kind == "s3iterabledataset":
```
muddyfish, line 65: If we're not targeting Python 3.10 minimum, can we at least change this to an if/elif block?

examples/benchmark/benchmark.py
```diff
@@ -0,0 +1,113 @@
...
+        dataset = S3IterableDataset.from_prefix(prefix_uri, region=region)
```
muddyfish, line 66: Would be nice here to allow us to mix/match `from_prefix` and `from_objects` as we're expecting them to have different performance levels

examples/benchmark/benchmark.py
```diff
@@ -0,0 +1,113 @@
...
+        dataset = torchdata.datapipes.iter.IterableWrapper(dataset)
+        if num_workers > 0:
+            dataset = dataset.sharding_filter()
+        if sharded:
+            dataset = dataset.map(tar_to_tuple)
+            dataset = dataset.load_from_tar()
+        return dataset.map(load_sample)
+    if kind == "s3mapdataset":
+        if sharded:
+            raise Exception(f"Sharding is not supported for {kind}")
```
muddyfish, line 76: Should this be a `NotImplementedError`? I don't like raising bare exceptions

examples/benchmark/benchmark_utils.py
```diff
@@ -0,0 +1,111 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  // SPDX-License-Identifier: BSD
+import time
+from dataclasses import dataclass
+
+import numpy as np
+import psutil
+import torch.cuda
+from pynvml import *
```
muddyfish, line 9: It's not good practice to do star imports. Can we just import exactly what we need?

examples/benchmark/README.md
vishalg0wda: Probably best. Will include it.

examples/benchmark/benchmark_utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+monitor_gpu = False
+if torch.cuda.is_available():
+    monitor_gpu = True
+    nvmlInit()
+
+
+@dataclass
+class ExperimentResult:
```
muddyfish, line 18: This class doesn't look like it's actually making use of the dataclass annotation?

examples/benchmark/benchmark_utils.py
```diff
@@ -0,0 +1,111 @@
...
+        self.num_samples = num_samples
+        self.throughput = self.num_samples / self.training_time
+        self.resource_data = {}
+        self.avg_resource_data = {}
+        self.checkpoint_results = checkpoint_results
+
+
+class ResourceMonitor:
+    """
+    Monitors cpu and gpu usage+memory.
```
muddyfish, line 32: CPU/GPU should be capitalised

examples/benchmark/benchmark_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    Set sleep_time_s carefully to avoid perf degradations.
+    """
+
+    def __init__(self, sleep_time_s=0.05, gpu_device=0, chunk_size=25_000):
```
muddyfish, line 36: Type hints?

examples/benchmark/README.md
vishalg0wda: Both are valid. If it's a string it expects it in either SI/IEC notation. The `datagen.py --help` has more information on the parameter formats. Do you think we should specify it in the README as well or will that suffice?

examples/benchmark/benchmark_utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def start(self):
+        self.monitor_thread = threading.Thread(target=self._monitor)
+        self.monitor_thread.start()
+
+    def stop(self):
+        self.stop_event.set()
+
+        for key, arr in self.resource_data.items():
+            self.resource_data[key] = arr[: self.cur_index]
```
muddyfish, line 87: Could we use `del arr[self.cur_index:]` here?

examples/benchmark/benchmark_utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+        self.monitor_thread.join()
```
muddyfish, line 89: Shouldn't we join before modifying `resource_data`?

examples/benchmark/benchmark_utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+        avgs = {}
+        for key, arr in self.resource_data.items():
+            last_vals = arr[start_index : self.cur_index]
+            avg = np.mean(last_vals)
+            avgs[key] = avg
+
+        return avgs
+
+    def get_avg_data(self):
```
muddyfish, line 105: These two functions are the same except one generates `start_index`. I think there should be a helper function that takes `start_index` in as a parameter

examples/benchmark/configuration/checkpointing.yaml
```diff
@@ -0,0 +1,14 @@
+defaults:
+  - dataloader: s3iterabledataset
+  - dataset: unsharded_dataset
+  - training: vit
+  - checkpoint: ???
+  - _self_
```
muddyfish, line 6: What does this mean?

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,205 @@
...
+        self.height = height
+        self.img_format = img_format
+
+    def generate(self, index: int, destdir: Path):
+        img = Image.fromarray(
+            np.random.randint(
+                0, high=256, size=(self.height, self.width, 3), dtype=np.uint8
+            )
+        )
+        img_file: Path = destdir / f"image_{index}.{self.img_format}"
```
muddyfish, line 39: Why are we saving these files to disk rather than using `BytesIO`?

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,205 @@
...
+        s3_client = boto3.client("s3", region_name=region)
+        s3_key_parts = []
+        if prefix:
+            s3_key_parts.append(prefix)
+        s3_key_parts.append(os.path.basename(file))
+        s3_key = "/".join(s3_key_parts)
+        s3_client.upload_file(file, bucket, s3_key)
+
+    @staticmethod
+    def parse_hr_bytes(ctx: click.Context, param: str, value: str) -> float | None:
```
muddyfish, line 90: What does `hr` stand for?

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,205 @@
...
+        if value is not None:
+            return prefixed.Float(value.rstrip("b").rstrip("B"))
+
+
+@click.command(context_settings={"show_default": True})
```
muddyfish, line 95: What advantages does click have over the built in argparse?

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,205 @@
...
+    resolution: tuple[int, int],
+    shard_size: float,
+    s3_bucket: str,
+    s3_prefix: str,
+    region: str,
+):
+    """
+    Synthesizes a dataset that will be used for benchmarking and uploads it to an S3 bucket.
+    """
+    with tempfile.TemporaryDirectory() as tempdir:
```
muddyfish, line 146: Can we please avoid writing to local storage if possible?

examples/benchmark/models.py
```diff
@@ -0,0 +1,194 @@
...
+        self.num_labels = num_labels
+
+    def load_sample(self, sample):
```
muddyfish, line 47: Type hint?

examples/benchmark/models.py
```diff
@@ -0,0 +1,194 @@
...
+        return self._transform
+
+    @property
+    def optimizer(self):
+        if not self._optimizer:
+            self._optimizer = torch.optim.Adam(self.model.parameters(), lr=0.001)
+        return self._optimizer
+
+    def load_sample(self, sample):
+        if isinstance(sample, S3Reader):
```
muddyfish, line 114: These lines are repeated a bit. Is it time to extract them?

examples/benchmark/models.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def train(
+        self, dataloader: torch.utils.data.dataloader, epochs: int
+    ) -> ExperimentResult:
+        num_samples = 0
+        start_time = time.perf_counter()
+        checkpoint_results = []
+        for epoch in range(0, epochs):
+            for batch_idx, (data, target) in enumerate(dataloader):
+                num_samples += len(data)
```
muddyfish, line 140: Could we move this block to it's own function?

examples/benchmark/models.py
```diff
@@ -0,0 +1,194 @@
...
+
+        return ExperimentResult(
+            training_time=training_time,
+            num_samples=num_samples,
+            checkpoint_results=checkpoint_results,
+        )
+
+    def save(self, batch_idx):
+        destination = self.checkpoint.destination
+        if destination == "s3":
```
muddyfish, line 165: A dict with str->function might be better here.

examples/benchmark/README.md
fuatbasik: Hah! because I wrote part of it :) I will start taking English classes.

examples/benchmark/benchmark.py
fuatbasik: Thats number of labels. To be fair, as we are working with synthetic data, any number is fine here. But of-course it would be better to make it parametrised.

examples/benchmark/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+```
+$ python datagen.py --help
+Usage: datagen.py [OPTIONS]
+
+  Synthesizes a dataset that will be used for benchmarking and uploads it to
+  an S3 bucket.
+
+Options:
+  -n, --num-samples FLOAT  Number of samples to generate.  [default: 1k]
```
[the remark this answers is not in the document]
vishalg0wda: We're using the [prefixed](https://pypi.org/project/prefixed/) package to translate SI/IEC prefixed numbers to numeric values. That package happens uses floats for the underlying type. I can see how that's confusing though.

examples/benchmark/README.md
```diff
@@ -0,0 +1,163 @@
...
+  --resolution TEXT        Resolution written in 'widthxheight' format
```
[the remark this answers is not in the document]
vishalg0wda: Probably subjective, but felt more natural to specify resolution in this format and as a tuple rather than individually.

examples/benchmark/README.md
```diff
@@ -0,0 +1,163 @@
...
+                           of the given size. Size can be supplied as an IEC
+                           or SI prefix. Eg: 16Mib, 4Kb, 1Gib.Note: these are
+                           case-sensitive notations.
+  --s3-bucket TEXT         S3 Bucket name. Note: Ensure the credentials are
+                           made available either through environment variables
+                           or a shared credentials file.  [required]
+  --s3-prefix TEXT         Optional S3 Key prefix where the dataset will be
+                           uploaded. Note: a prefix will be autogenerated. eg:
+                           s3://<BUCKET>/1k_256x256_16Mib_sharded/
+  --region TEXT            Region where the S3 bucket is hosted.  [default:
```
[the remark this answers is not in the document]
vishalg0wda: There should be. It must've spilt over to the next line.

examples/benchmark/benchmark.py
vishalg0wda: Agree with the comment re bare exceptions. I don't think it will ever be supported either either no? I'm thinking a `ValueError` would be more appropriate.

examples/benchmark/benchmark_utils.py
vishalg0wda: Good spot! Yes, we should.

examples/benchmark/benchmark_utils.py
vishalg0wda: Is this suggestion for readbility or is it a performance concern? 

`del` doesn't work against numpy arrays, but `numpy.delete` does. The approach above should be equivalent in performance to delete though since these are both ["no-copy" operations](https://towardsdatascience.com/the-performance-advantage-of-no-copy-dataframe-operations-7bf8c565c9a0#:~:text=When%20you%20slice%20a%20NumPy,and%20strides%20into%20that%20data.).

examples/benchmark/configuration/checkpointing.yaml
vishalg0wda: It controls the override behavior, though looking at [the docs](https://hydra.cc/docs/advanced/defaults_list/#composition-order), it appears adding `_self_` at the bottom is the default behavior making this redundant. I'll omit this in the next revision.

examples/benchmark/datagen.py
vishalg0wda: The original version was written to synthesize files on disk and rely on `aws s3 sync` which would handle the upload externally. 

Since we're handling uploads as well now, we should ideally be implementing it as a a pipeline like what you're suggesting(especially if datasets are to get really large).

I'm thinking it should be reasonable to address this in a future iteration and take this down as a planned enhancement. Let me know what you think.

examples/benchmark/datagen.py
vishalg0wda: Sorry for the obscure shortform. `hr == human readable` (oh the irony!)

examples/benchmark/benchmark_utils.py
muddyfish: It wasn't a performance concern, more a readability one. It took a couple of seconds to work out what was being done here, and I think a mutable delete is more readable here. What do you think?

examples/benchmark/benchmark.py
muddyfish: Using a `ValueError` sounds right to me

examples/benchmark/README.md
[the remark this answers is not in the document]
muddyfish: Is there a way to tell the package we only want to handle integers here, and floats are invalid?

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,345 @@
...
+                total = item_size
+            else:
+                group.append(item)
+                total = new_total
+        if len(group) > 0:
+            yield group
+
+    @staticmethod
+    def get_sample_size(item: tuple[str, io.BytesIO]) -> int:
+        nbytes = item[1].getbuffer().nbytes
```
muddyfish, line 81: Why not just return this in one line? I don't think the variable name is helping here

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,345 @@
...
+        pass
+
+
+class ImageGenerator(DataGenerator):
+    def __init__(self, width: int, height: int, img_format: str = "JPEG"):
+        self.width = width
+        self.height = height
+        self.img_format = img_format
+
+    def create(self, idx_iter: Iterator[int]) -> Iterator[tuple[str, io.BytesIO]]:
```
muddyfish, line 32: We're using `tuple[str, BytesIO]` a few times here. Why not create a named tuple for it, or a dataclass?

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,345 @@
...
+    @staticmethod
+    def parse_human_readable_bytes(
+        ctx: click.Context, param: str, value: str
+    ) -> float | None:
+        if value is not None:
+            return pr.Float(value.rstrip("b").rstrip("B"))
+
+    @classmethod
+    def write_dataset_config(cls, disambiguator: str, dataset_cfg: dict[str, Any]):
+        cfg_path = f"./configuration/dataset/{disambiguator}.yaml"
```
muddyfish, line 123: Is this really what we want, or are we wanting to write relative to the actual script path? Ideally we'd also use `os.path.join` here

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,345 @@
...
+Sentinel = NewType("Sentinel", None)
+
+
+class ThreadSafeIterator:
+    """Takes an iterator/generator and makes it thread-safe by
+    serializing call to the `next` method of given iterator/generator.
+    """
+
+    def __init__(self, it):
+        self.it = it
```
muddyfish, line 138: Nit: These attributes should be private

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,345 @@
...
+            return self.it.__next__()
+
+
+def build_pipeline(
+    num_samples: float, resolution: tuple[int, int], shard_size: float
+) -> Iterator[tuple[str, io.BytesIO]]:
+    # define the pipeline to generate the dataset
+    # TODO: parameterize the data generator to allow for creating other kinds of datasets(eg: text).
+    sample_generator = ImageGenerator(*resolution)
+    pipeline: Iterator[tuple[str, io.BytesIO]] = sample_generator.create(
```
muddyfish, line 155: Nit: no need for a generator comprehension here. Just using the range function directly should work

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,345 @@
...
+        )
+
+    return ThreadSafeIterator(pipeline)
+
+
+def build_producers(
+    num_workers: int, queue: Queue, dataset_generator: Iterator[tuple[str, io.BytesIO]]
+) -> list[Thread]:
+    barrier = Barrier(num_workers)
+    out = []
```
muddyfish, line 176: This is just a list comprehension like the function below. Can we be consistent between the two? Either by changing one or the other

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,345 @@
...
+    shard_size: float,
+    s3_bucket: str,
+    s3_prefix: str,
+    region: str,
+):
+    """
+    Synthesizes a dataset that will be used for benchmarking and uploads it to an S3 bucket.
+    """
+    num_workers = os.cpu_count()
+    Q = Queue(num_workers)
```
muddyfish, line 291: Variable names are lowercase. This is also not descriptive

examples/benchmark/models.py
```diff
@@ -0,0 +1,197 @@
...
+                    v2.ToImage(),
+                    v2.ToDtype(torch.uint8, scale=True),
+                    v2.RandomResizedCrop((224, 224), antialias=True),
+                    v2.ToDtype(torch.float32, scale=True),
+                    v2.Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225]),
+                ]
+            )
+        return self._transform
+
+    @property
```
muddyfish, line 112: cached_property should work here better

muddyfish:
I'm also planning on testing this out with this PR later: https://github.com/awslabs/s3-connector-for-pytorch/pull/145

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,345 @@
...
+            return self.val
+
+
+class Utils:
+    T = TypeVar("T")
+
+    @staticmethod
+    def batcher(
+        items: Iterator[T], size_extractor: Callable[[T], int], size_threshold: float
+    ) -> Iterator[list[T]]:
```
muddyfish, line 63: Similarly, we can either use `List[]` or do a future annotations import for py3.8 support

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,345 @@
...
+
+    @classmethod
+    def write_dataset_config(cls, disambiguator: str, dataset_cfg: dict[str, Any]):
+        cfg_path = f"./configuration/dataset/{disambiguator}.yaml"
+        with open(cfg_path, "w") as outfile:
+            yaml.dump(dataset_cfg, outfile, default_flow_style=False)
+            click.echo(f"Dataset Configuration created at: {cfg_path}")
+
+
+Sentinel = NewType("Sentinel", None)
```
muddyfish, line 129: In py3.8, there's a pickling error relating to `NewType`. When you add unit tests, can you make sure all our supported Python versions are tested properly?

examples/benchmark/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+_Note: For overriding any other benchmark parameters,
+see [Hydra Overrides](https://hydra.cc/docs/advanced/override_grammar/basic/)._
+
+Experiments will report total training time, number of training samples as well as host-level metrics like CPU
+Utilisation, GPU Utilisation (if available) etc.
+
+## Next Steps
+
+- Generate and upload samples directly to S3 bypassing the filesystem.
```
muddyfish, line 175: This is done right?

examples/benchmark/benchmark.py
```diff
@@ -0,0 +1,128 @@
...
+    )
+
+    print(
+        f"Resource usage of the workload was as follows:\n"
+        f"{result.avg_resource_data}"
+    )
+
+
+def make_model(config: DictConfig) -> ModelInterface:
+    if config.training.model == "entitlement":
```
muddyfish, line 54: Not blocking: It would be great to have this being automatic, with classes auto-registering themselves to a model dictionary. When creating a new model, editing this manually was a little annoying.

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,354 @@
...
+        with self._lock:
+            return self._it.__next__()
+
+
+def build_pipeline(
+    num_samples: float, resolution: Tuple[int, int], shard_size: float
+) -> Iterator[LabelledSample]:
+    # define the pipeline to generate the dataset
+    # TODO: parameterize the data generator to allow for creating other kinds of datasets(eg: text).
+    sample_generator = ImageGenerator(*resolution)
```
muddyfish, line 166: Would be nice to be able to include the file format as an option here.

examples/benchmark/datagen.py
```diff
@@ -0,0 +1,354 @@
...
+    # signal that there are no further items
+    if identifier == 0:
+        queue.put(Sentinel())
+
+
+def consumer(queue: Queue, activity: Callable[[LabelledSample], None], identifier: int):
+    while True:
+        # click.echo(f"Consumer running on thread {threading.current_thread().ident}")
+        item: Union[LabelledSample, Sentinel] = queue.get()
+        if type(item) is Sentinel:
```
muddyfish, line 239: Nit: Is there any reason we couldn't put in the Sentinel class itself in, and then just `if item is Sentinel`?

examples/benchmark/models.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+class Entitlement(ModelInterface):
+    """
+    This is not really a training model as it does not train anything. Instead, this model simply reads the binary
+    object data from S3, so that we may identify the max achievable throughput for a given dataset.
+    """
+
+    def __init__(self, num_labels: int = None):
+        super(Entitlement, self).__init__("Entitlement")
```
muddyfish, line 73: The name attribute here is always the classes name. Why not use `type(self).__name__` in the parent/usages?

s3torchbenchmarking/README.md
```diff
@@ -0,0 +1,187 @@
...
+    # Example-2:
+    $ s3torch-benchmark -cd conf -m -cn checkpointing 'dataset.prefix_uri=<S3-PREFIX>' 'dataset.region=eu-west-2'
+
+_Note: For overriding any other benchmark parameters,
+see [Hydra Overrides](https://hydra.cc/docs/advanced/override_grammar/basic/). You can also run `s3torch-benchmark --hydra-help` to learn more._
+
+Experiments will report total training time, number of training samples as well as host-level metrics like CPU
+Utilisation, GPU Utilisation (if available) etc.
+
+## Next Steps
```
muddyfish, line 183: Not blocking: It would be great to include instructions on how to configure logging, especially the log level.

s3torchbenchmarking/pyproject.toml
```diff
@@ -0,0 +1,32 @@
+[build-system]
+requires = ["setuptools", "build"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "s3torchbenchmarking"
+version = "0.0.1"
+description = "Tools to run and compare benchmarks against various PyTorch connectors like the s3torchconnector."
+requires-python = ">=3.8,<3.12"
```
muddyfish, line 9: Does this explicitly require less than 3.12? Ideally I think we'd want to make this as lenient as possible, and take our version restriction from s3torchconnector itself

s3torchbenchmarking/pyproject.toml
```diff
@@ -0,0 +1,32 @@
+[build-system]
+requires = ["setuptools", "build"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "s3torchbenchmarking"
+version = "0.0.1"
+description = "Tools to run and compare benchmarks against various PyTorch connectors like the s3torchconnector."
```
muddyfish, line 8: Nit: `like s3torchconnector`

s3torchbenchmarking/src/s3torchbenchmarking/datagen.py
```diff
@@ -0,0 +1,380 @@
...
+        self.val = 0
+        self.lock = Lock()
+
+    def get_and_inc(self):
+        with self.lock:
+            self.val += 1
+            return self.val
+
+
+class Utils:
```
muddyfish, line 71: Not blocking: Why is this in a class? Most (all?) of these functions would work perfectly well at the top level

s3torchbenchmarking/tst/test_compatibility.py
```diff
@@ -0,0 +1,5 @@
+def test_imports():
```
muddyfish, line 1: Not blocking: This is a good start for tests, but in the future it would be great to include more tests to ensure the functionality is working properly

muddyfish:
Thanks for making all the changes, it looks a lot better now. Everything else I've written is not blocking, and I'm happy to approve this once the tests pass

muddyfish (approved):

dnanuti (approved):

PR #139, opened 2024-02-12
WIP: Add integration tests for Lightning support

dnanuti:
## Description

This PR adds E2E tests for Lightning checkpoints. CI tasks and imports need to be double checked.
## Related items

- https://github.com/awslabs/s3-connector-for-pytorch/pull/132

## Testing

- Locally ran against personal S3 bucket
--------

### DIFF
--- a/.github/workflows/rust-checks.yml
+++ b/.github/workflows/rust-checks.yml
@@ -37,7 +37,8 @@ jobs:
       - name: Run cargo deny
         uses: EmbarkStudios/cargo-deny-action@v1
         with:
-          command: --manifest-path s3torchconnectorclient/Cargo.toml check ${{ matrix.checks }}
+          command: check ${{ matrix.checks }}
+          manifest-path: s3torchconnectorclient/Cargo.toml
 
   clippy:
     runs-on: ubuntu-22.04
--- a/s3torchconnectorclient/Cargo.lock
+++ b/s3torchconnectorclient/Cargo.lock
@@ -4,13 +4,14 @@ version = 3
 
 [[package]]
 name = "ahash"
-version = "0.7.7"
+version = "0.8.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5a824f2aa7e75a0c98c5a504fceb80649e9c35265d44525b5f94de4771a395cd"
+checksum = "42cd52102d3df161c77a887b608d7a4897d7cc112886a9537b738a887a03aaff"
 dependencies = [
- "getrandom",
+ "cfg-if",
  "once_cell",
  "version_check",
+ "zerocopy",
 ]
 
 [[package]]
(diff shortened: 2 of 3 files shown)


s3torchconnector/tst/e2e/models/lightning_transformer.py
```diff
@@ -0,0 +1,26 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  // SPDX-License-Identifier: BSD
+from typing import Dict, Any
+
+import lightning as L
```
muddyfish, line 5: Why not do a from import LightningModule here given that we're only using this once?

s3torchconnector/tst/e2e/models/net.py
```diff
@@ -0,0 +1,31 @@
+import torch
```
muddyfish, line 1: Missing copyright. Though are these files something Amazon owns, or did you copy/paste this from an example?

s3torchconnector/tst/e2e/test_e2e_s3_lightning_checkpoint.py
```diff
@@ -0,0 +1,102 @@
...
+    dataloader = DataLoader(dataset)
+    model = LightningTransformer(vocab_size=dataset.vocab_size)
+    trainer = L.Trainer(fast_dev_run=3)
+    trainer.fit(model=model, train_dataloaders=dataloader)
+    checkpoint_name = "lightning_module_training_checkpoint.pt"
+    s3_uri = f"{checkpoint_directory.s3_uri}/{checkpoint_name}"
+    trainer.save_checkpoint(s3_uri)
+    lightning_checkpoint = S3LightningCheckpoint(region=checkpoint_directory.region)
+    loaded_model = lightning_checkpoint.load_checkpoint(s3_uri)
+    for model_item, loaded_item in zip(
```
muddyfish, line 61: Do you think `(model_key, model_value), (loaded_key, loaded_item) in ...` would look better, and then you'd not have to do the index lookups later

s3torchconnector/tst/e2e/models/net.py
dnanuti: This was previously added as part of the checkpoint work, but I think it's from an example too - https://pytorch.org/tutorials/beginner/blitz/neural_networks_tutorial.html, that's why I removed the copyright, but missed removing it in my class.

IsaevIlya (commented):
Could we cover a few more cases, please:
- case for using `S3LightningCheckpoint` wrapped in `AsyncCheckpointIO`
- one more case for usage as plugin with callback to test `save` and `delete` path at the same time:
```
checkpoint_callback = ModelCheckpoint(
    dirpath="s3://bucket_name/key_prefix/", 
    save_top_k=1,
    every_n_epochs=1,
    monitor="val_loss",
    filename="checkpoint-{epoch:02d}-{val_loss:.2f}.ckpt"
)
trainer = Trainer(
    plugins=[s3_checkpoint_io],
    callbacks=[checkpoint_callback]
)
trainer.fit(model)
# after finishing there should be only one checkpoint left
# name of the checkpoint should be the same as checkpoint_callback.best_model_path
```
- case for loading data from checkpoint via plugin:
```
trainer = pl.Trainer(
    plugins=[s3_checkpoint_io], 
    resume_from_checkpoint='s3://bucket_name/key_prefix/checkpoint.ckpt'
)
```

dnanuti:
This was automatically closed due to a bad rebase.

PR #143, opened 2024-02-13
Add integration tests for Lightning support

dnanuti:
## Description

This PR adds E2E tests for Lightning checkpoints.
## Related items

- https://github.com/awslabs/s3-connector-for-pytorch/pull/132

## Testing

- Locally ran against personal S3 bucket
--------

### DIFF
--- a/s3torchconnector/tst/e2e/test_e2e_s3_lightning_checkpoint.py
+++ b/s3torchconnector/tst/e2e/test_e2e_s3_lightning_checkpoint.py
@@ -1,0 +1,217 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  // SPDX-License-Identifier: BSD
+import pytest
+import random
+import torch
+
+from pathlib import Path
+from typing import Dict, Any
+from lightning.pytorch.callbacks import ModelCheckpoint
+from lightning.pytorch.demos import WikiText2
+from lightning.pytorch.plugins import AsyncCheckpointIO
+from torch.utils.data import DataLoader
+
+from s3torchconnector import S3Checkpoint
+from s3torchconnector._s3dataset_common import parse_s3_uri
+from s3torchconnector.lightning import S3LightningCheckpoint
+from s3torchconnectorclient import S3Exception
+
+from models.net import Net
+from models.lightning_transformer import LightningTransformer, L
+
+
+def test_save_and_load_checkpoint(checkpoint_directory):
+    tensor = torch.rand(3, 10, 10)
+    s3_lightning_checkpoint = S3LightningCheckpoint(region=checkpoint_directory.region)
+    checkpoint_name = "lightning_checkpoint.ckpt"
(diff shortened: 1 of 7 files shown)


IsaevIlya (commented):
Could we cover a few more cases, please:

- case for using S3LightningCheckpoint wrapped in AsyncCheckpointIO
- one more case for usage as plugin with callback to test save and delete path at the same time:
```
checkpoint_callback = ModelCheckpoint(
    dirpath="s3://bucket_name/key_prefix/", 
    save_top_k=1,
    every_n_epochs=1,
    monitor="val_loss",
    filename="checkpoint-{epoch:02d}-{val_loss:.2f}.ckpt"
)
trainer = Trainer(
    plugins=[s3_checkpoint_io],
    callbacks=[checkpoint_callback]
)
trainer.fit(model)
# after finishing there should be only one checkpoint left
# name of the checkpoint should be the same as checkpoint_callback.best_model_path
```
- case for loading data from checkpoint via plugin:

```
trainer = pl.Trainer(
    plugins=[s3_checkpoint_io], 
    resume_from_checkpoint='s3://bucket_name/key_prefix/checkpoint.ckpt'
)
```

s3torchconnector/tst/e2e/test_e2e_s3_lightning_checkpoint.py
```diff
@@ -0,0 +1,217 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  // SPDX-License-Identifier: BSD
+import pytest
+import random
+import torch
+
+from pathlib import Path
+from typing import Dict, Any
+from lightning.pytorch.callbacks import ModelCheckpoint
```
IsaevIlya, line 9: We may need to add dependency on https://pypi.org/project/pytorch-lightning/ to have access to `ModelCheckpoint`

s3torchconnector/tst/e2e/test_e2e_s3_lightning_checkpoint.py
```diff
@@ -0,0 +1,217 @@
...
+        min_epochs=4,
+        max_epochs=5,
+        max_steps=3,
+    )
+
+    trainer.fit(model, dataloader)
+    expected_checkpoint_name = "checkpoint-epoch=00-step=03.ckpt"
+    # Workaround to verify there is only one checkpoint file. Client should NOT be accessed this way.
+    bucket, prefix = parse_s3_uri(checkpoint_directory.s3_uri)
+    list_result = list(s3_lightning_checkpoint._client.list_objects(bucket, prefix))
```
IsaevIlya, line 135: Is there any reason why we are going to use approach that we explicitly call a "bad" one ourself?

dnanuti: Good catch, forgot the client is available there 😄

IsaevIlya (approved):

Issue #133, opened 2024-02-09
Extra HEAD request is made when using `from_prefix` constructor


muddyfish:
### s3torchconnector version

s3torchconnector-1.1.3 (from source)

### s3torchconnectorclient version

s3torchconnectorclient-1.1.3 (from source)

### AWS Region

eu-west-1

### Describe the running environment

Intel Mac

### What happened?

Using the `from_prefix` constructor for datasets is meant to avoid using HEAD requests when seeking to the end, but it doesn't.

Relevant logs:
```
DEBUG s3torchconnector._s3client._s3client 2024-02-09 11:14:43,448 _s3client.py:90 - GetObjectWithInfo s3://s3torchconnector-demo/geonet/images/DISC/DISC.01/2022.001/2022.001.0000.00.DISC.01.jpg
DEBUG s3torchconnector._s3client._s3client 2024-02-09 11:14:43,448 _s3client.py:58 - GetObject s3://s3torchconnector-demo/geonet/images/DISC/DISC.01/2022.001/2022.001.0000.00.DISC.01.jpg
DEBUG s3torchconnector._s3client._s3client 2024-02-09 11:14:43,448 _s3client.py:84 - HeadObject s3://s3torchconnector-demo/geonet/images/DISC/DISC.01/2022.001/2022.001.0000.00.DISC.01.jpg
```

```
s3_dataset = s3torchconnector.S3IterableDataset.from_prefix(SHARDS_URI, region=REGION)

for f_obj in s3_dataset:
    f_obj.seek(0, SEEK_END)
    break
```

### Relevant log output

```shell
INFO s3torchconnector.s3iterable_dataset 2024-02-09 11:14:42,724 s3iterable_dataset.py:90 - Building S3IterableDataset from_prefix s3_uri='s3://s3torchconnector-demo/geonet/images/DISC/DISC.01/2022.001/'
DEBUG s3torchconnector._s3client._s3client 2024-02-09 11:14:42,724 _s3client.py:79 - ListObjects s3://s3torchconnector-demo/geonet/images/DISC/DISC.01/2022.001/
TRACE mountpoint_s3_client.s3_crt_client 2024-02-09 11:14:42,736 s3_crt_client.rs:253 - constructing client with auth config Default
[...]

PR #145, opened 2024-02-14
Remove extra HeadObject call in from_prefix

muddyfish:
## Description

- Ensure HeadObject is not explicitly called when seeking to the end of a file when dataset is created with from_prefix
- Rename `S3BucketKey` to `S3BucketKeyData`
- Change signature of `iter(S3BucketIterable)` to `Iterator[S3BucketData]` instead of `Iterator[S3Reader]`

## Additional context

No breaking changes

## Related items

https://github.com/awslabs/s3-connector-for-pytorch/issues/133

## Testing

Extra tests added which verify that `head_object` is not called when seeking to the end of an S3Object when created using `from_prefix` for either dataset.

--------

### DIFF
--- a/s3torchconnector/src/s3torchconnector/_s3_bucket_iterable.py
+++ b/s3torchconnector/src/s3torchconnector/_s3_bucket_iterable.py
@@ -11,8 +11,8 @@ from s3torchconnectorclient._mountpoint_s3_client import (
     ListObjectStream,
 )
 
+from ._s3bucket_key_data import S3BucketKeyData
 from ._s3client import S3Client
-from . import S3Reader
 
 
 class S3BucketIterable:
@@ -21,7 +21,7 @@ class S3BucketIterable:
         self._bucket = bucket
         self._prefix = prefix
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator[S3BucketKeyData]:
         # This allows us to iterate multiple times by re-creating the `_list_stream`
         return iter(S3BucketIterator(self._client, self._bucket, self._prefix))
 
@@ -32,10 +32,9 @@ class S3BucketIterator:
         self._bucket = bucket
         self._list_stream = _PickleableListObjectStream(client, bucket, prefix)
 
-    def __iter__(self) -> Iterator[S3Reader]:
-        return map(
-            partial(self._client.from_bucket_and_object_info, self._bucket),
-            chain.from_iterable(map(_extract_object_info, self._list_stream)),
+    def __iter__(self) -> Iterator[S3BucketKeyData]:
+        return chain.from_iterable(
+            map(partial(_extract_list_results, self._bucket), self._list_stream)
         )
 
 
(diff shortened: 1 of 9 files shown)


muddyfish:
I'm planning on running some benchmarking on this to validate this has performance improvements in real-world usage

IsaevIlya (approved):

s3torchconnector/src/s3torchconnector/_s3client/_s3client.py
```diff
@@ -54,12 +54,19 @@ def _client_builder(self) -> MountpointS3Client:
...
         )
 
-    def get_object(self, bucket: str, key: str) -> S3Reader:
-        log.debug(f"GetObject s3://{bucket}/{key}")
+    def get_object(
+        self, bucket: str, key: str, *, object_info: Optional[ObjectInfo] = None
+    ) -> S3Reader:
+        log.debug(f"GetObject s3://{bucket}/{key}, {object_info is None=}")
+        if object_info is None:
+            get_object_info = partial(self.head_object, bucket, key)
```
dnanuti, line 62: Just a thought, should we include debug logs in the if branches as well?

s3torchconnector/src/s3torchconnector/_s3_bucket_iterable.py
```diff
@@ -32,10 +32,9 @@ def __init__(self, client: S3Client, bucket: str, prefix: str):
         self._bucket = bucket
         self._list_stream = _PickleableListObjectStream(client, bucket, prefix)
 
-    def __iter__(self) -> Iterator[S3Reader]:
-        return map(
-            partial(self._client.from_bucket_and_object_info, self._bucket),
-            chain.from_iterable(map(_extract_object_info, self._list_stream)),
+    def __iter__(self) -> Iterator[S3BucketKeyData]:
```
dnanuti, line 35: I understand why we need this change, but at the first glance it looks like we're creating a stronger coupling with S3BucketIterable, which we plan to simplify at some point. Is it worth to invest in that simplification now, rather than creating more dependencies on it?

dnanuti (changes requested):

s3torchconnector/src/s3torchconnector/_s3client/_s3client.py
muddyfish: What additional stuff would we log? I'm already logging if it takes the branch or not with the `object_info is None` logging

s3torchconnector/src/s3torchconnector/_s3_bucket_iterable.py
muddyfish: I think changing the signature here is fine, given that it's only used in one place and is a private class. Also, the previous usages were only to unpack the key/bucket out of the `S3Reader` so this requires less packing/unpacking. 

I do agree that we should consider simplifying this code at some point, though I'm not sure if this PR is the best time for it

s3torchconnector/src/s3torchconnector/_s3client/_s3client.py
dnanuti: Missed that, thanks!

dnanuti (dismissed):

PR #150, opened 2024-02-20
Add examples for reading/writing checkpoints with lightning in various ways

muddyfish:
## Description

Adds four example files for using S3 connector for Pytorch with Pytorch Lightning:

- Reading from a checkpoint
- Writing to checkpoints automatically, including options to keep the most recent checkpoint only.
- Manually writing to a checkpoint
- Asynchronous checkpoint writing

## Additional context

No breaking changes

## Testing

Ran all files, and saw they wrote files, and read files from S3.

--------

### DIFF
--- a/examples/lightning/async_checkpoint_writing.py
+++ b/examples/lightning/async_checkpoint_writing.py
@@ -1,0 +1,44 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  // SPDX-License-Identifier: BSD
+
+from lightning import Trainer
+from lightning.pytorch.callbacks import ModelCheckpoint
+from lightning.pytorch.demos import WikiText2, LightningTransformer
+from lightning.pytorch.plugins import AsyncCheckpointIO
+from torch.utils.data import DataLoader
+
+from s3torchconnector.lightning import S3LightningCheckpoint
+
+
+def main(region: str, checkpoint_path: str):
+    dataset = WikiText2()
+    dataloader = DataLoader(dataset, num_workers=3)
+
+    model = LightningTransformer(vocab_size=dataset.vocab_size)
+    s3_lightning_checkpoint = S3LightningCheckpoint(region)
+    async_checkpoint = AsyncCheckpointIO(s3_lightning_checkpoint)
+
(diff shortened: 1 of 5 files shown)


IsaevIlya (commented):
Could we please add the small section about using S3LightningCheckpoint into Readme.md. That we are providing support for Lightning checkpoints, how to install that module, simple example of usage like:
```
trainer = Trainer(
    plugins=[s3_checkpoint_io],
    default_root_dir="s3://bucket_name/key_prefix/"
)
trainer.fit(model)
```

examples/lightning/async_checkpoint_writing.py
```diff
@@ -0,0 +1,42 @@
...
+
+def main(region: str, checkpoint_path: str):
+    dataset = WikiText2()
+    dataloader = DataLoader(dataset, num_workers=3)
+
+    model = LightningTransformer(vocab_size=dataset.vocab_size)
+    s3_lightning_checkpoint = S3LightningCheckpoint(region)
+    async_checkpoint = AsyncCheckpointIO(s3_lightning_checkpoint)
+
+    checkpoint_callback = ModelCheckpoint(
```
dnanuti, line 21: Can we perhaps add some explanatory comments such as:
- checkpoint path will look like ... 
- we expect N checkpoints to be saved ... etc.

Same for the other examples.

examples/lightning/checkpoint_reading.py
```diff
@@ -0,0 +1,31 @@
...
+    model = LightningTransformer(vocab_size=dataset.vocab_size)
+    s3_lightning_checkpoint = S3LightningCheckpoint(region)
+
+    trainer = Trainer(
+        plugins=[s3_lightning_checkpoint],
+        min_epochs=4,
+        max_epochs=5,
+        max_steps=3,
+    )
+    # Load the checkpoint in `ckpt_path` before training
```
dnanuti, line 24: nit: load in CHECKPOINT_PATH

dnanuti (approved):

examples/lightning/checkpoint_reading.py
muddyfish: I was thinking it makes more sense to use `ckpt_path` as that's the name of the argument as supplied by Lightning.

PR #153, opened 2024-02-21
Add ARM runners to the CI

muddyfish:
## Description

- Automatically build ARM wheels in our CI with codebuild
- Also migrate x86 wheel builds to codebuild
- Total wheel build improved from 20-30 minutes to ~15 minutes

## Additional context

No breaking changes

## Testing

Passed workflow: https://github.com/awslabs/s3-connector-for-pytorch/actions/runs/7991821909
Wheels artifact includes x86, arm, and source wheels.

--------

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -18,17 +18,22 @@ jobs:
     uses: ./.github/workflows/generate_third_party_licenses.yml
 
   build_wheels:
-    name: Build wheels for ${{ matrix.build_target }}
-    runs-on: ubuntu-20.04
+    name: Build wheels for ${{ matrix.build_prefix }} - ${{ matrix.builder.arch }}
+    runs-on: ${{ matrix.builder.runner }}
     needs: generate_third_party_licenses
     strategy:
       matrix:
-        build_target:
-          - cp38-manylinux_x86_64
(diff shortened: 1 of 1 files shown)


.github/workflows/wheels.yml
dnanuti, line 1: Just a general comment, we should probably stick to the best practices for this:

```

Build wheels for cp310 - x86_64
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
```

dnanuti: source: https://github.com/awslabs/s3-connector-for-pytorch/actions/runs/7991821909

muddyfish: I looked into this, and the place the error issues from is inside `actions/setup-python`. It looks like the action assumes we're running inside a container. I don't think this is the worst issue in the world with regards to security issues either, as in case of problems, the user will already have access to our wheel environment and aws credentials. I don't think a Python virtual environment is likely to include much in the way of protection

https://github.com/awslabs/s3-connector-for-pytorch/actions/runs/7992488276/job/21826211484

dnanuti (dismissed):

PR #157, opened 2024-02-26
Support benchmarking mountpoint-backed datasets

vishalg0wda:
## Description
Adds support for running benchmarks against mountpoints-backed datasets. The feature relies on the `mount-s3` binary being installed and relies on the `FSSpecFileLister` and `FSSpecFileOpener` APIs exposed by torchdata to interact with the filesystem. A reference configuration `dataloader/mountpoint.yaml` has already been created that can be used in experiments. 

Note: This feature mounts all the objects under an s3 prefix at a temporary path on the filesystem, and also manages the clean up of the same via a signal handler that is invoked when the process terminates.

## Testing
Locally tested a sample scenario by running:
```
AWS_PROFILE=<PROFILE-NAME> ./venv/bin/s3torch-benchmark -cd conf -m -cn dataloading dataset.prefix_uri=<S3-PREFIX> dataloader=mountpoint dataset.region=<AWS-REGION>
```

```
2024-02-26 13:12:23,150][HYDRA] Launching 4 jobs locally
[2024-02-26 13:12:23,150][HYDRA] 	#0 : dataloader=mountpoint dataloader.num_workers=2 <redacted>
bucket swift-benchmark-dataset is mounted at /tmp/s3dataset_845qhz65
mountpoint trained entitlement in 33.7384s with 2.9640 samples per second
Resource usage of the workload was as follows:
{'cpu_util': 11.367466266866565, 'cpu_mem': 37.06821589205397, 'gpu_util': 0.0, 'gpu_mem': 0.0}
[2024-02-26 13:13:02,394][HYDRA] 	#1 : dataloader=mountpoint dataloader.num_workers=4 <redacted>
bucket swift-benchmark-dataset is mounted at /tmp/s3dataset_j464lkfh
mountpoint trained entitlement in 20.6110s with 4.8518 samples per second
Resource usage of the workload was as follows:
{'cpu_util': 15.406617647058823, 'cpu_mem': 37.098039215686285, 'gpu_util': 0.0, 'gpu_mem': 0.0}
[2024-02-26 13:13:24,478][HYDRA] 	#2 : dataloader=mountpoint dataloader.num_workers=8 <redacted>
bucket swift-benchmark-dataset is mounted at /tmp/s3dataset_bcwbo7h9
mountpoint trained entitlement in 18.9471s with 5.2779 samples per second
Resource usage of the workload was as follows:
{'cpu_util': 18.729066666666665, 'cpu_mem': 37.28906666666666, 'gpu_util': 0.0, 'gpu_mem': 0.0}
[2024-02-26 13:13:44,917][HYDRA] 	#3 : dataloader=mountpoint dataloader.num_workers=16 <redacted>
bucket swift-benchmark-dataset is mounted at /tmp/s3dataset_4lxrb1f4
/home/ANT.AMAZON.COM/vgd/w/s3-connector-for-pytorch/s3torchbenchmarking/venv/lib/python3.8/site-packages/torch/utils/data/dataloader.py:558: UserWarning: This DataLoader will create 16 worker processes in total. Our suggested max number of worker in current system is 8, which is smaller than what this DataLoader is going to create. Please be aware that excessive worker creation might get DataLoader running slow or even freeze, lower the worker number to avoid potential slowness/freeze if necessary.
  warnings.warn(_create_warning_msg(
mountpoint trained entitlement in 20.0413s with 4.9897 samples per second
Resource usage of the workload was as follows:
{'cpu_util': 23.0896725440806, 'cpu_mem': 37.688161209068014, 'gpu_util': 0.0, 'gpu_mem': 0.0}

```

--------

### DIFF
--- a/s3torchbenchmarking/README.md
+++ b/s3torchbenchmarking/README.md
@@ -75,6 +75,17 @@ Then from this directory, install the dependencies:
 This would make the `s3torch-benchmark` and `s3torch-datagen` commands available to you. Note: the installation would
 recommend $PATH modifications if necessary, allowing you to use the commands directly.
 
+**(Optional) Install Mountpoint**
+
+Required only if you're running benchmarks using PyTorch with Mountpoint for S3.
+
+    wget https://s3.amazonaws.com/mountpoint-s3-release/latest/x86_64/mount-s3.rpm
+    sudo yum install ./mount-s3.rpm # For an RHEL system
+
+For other distros see [Installing Mountpoint for Amazon S3](https://github.com/awslabs/mountpoint-s3/blob/main/doc/INSTALL.md).
+
+_Note: Mountpoint benchmarks are currently only supported on *nix-based systems and rely on `sudo` capabilities._  
+
 ### (Pre-requisite) Configure AWS Credentials
 
 The commands provided below(`datagen.py`, `benchmark.py`) rely on the standard [AWS credential discovery mechanism](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html). 
--- a/s3torchbenchmarking/src/s3torchbenchmarking/benchmark.py
+++ b/s3torchbenchmarking/src/s3torchbenchmarking/benchmark.py
@@ -1,19 +1,23 @@
 #  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #  // SPDX-License-Identifier: BSD
-
-
+import atexit
+import shutil
+import subprocess
+import tempfile
 from enum import Enum
-from typing import Optional
+from pathlib import Path
+from typing import Optional, List
 
 import hydra
 import torchdata
 from omegaconf import DictConfig
-from s3torchconnector import S3IterableDataset, S3Reader, S3MapDataset
 from torch.utils.data import DataLoader, Dataset, default_collate
 from torchdata.datapipes.utils import StreamWrapper
 
 from s3torchbenchmarking.benchmark_utils import ResourceMonitor
 from s3torchbenchmarking.models import Entitlement, ViT, ModelInterface
+from s3torchconnector import S3IterableDataset, S3Reader, S3MapDataset
+from s3torchconnector._s3dataset_common import parse_s3_uri
 
 
 @hydra.main(version_base=None)
@@ -71,6 +75,27 @@ class DatasetSharding(Enum):
             return DatasetSharding[dataset_config.sharding]
 
 
+def make_mountpoint(
+    prefix_uri: str,
+    mountpoint_path: Optional[str] = None,
+    additional_args: List[str] = None,
+) -> str:
+    def teardown(path: str):
+        subprocess.run(["sudo", "umount", path])
+        shutil.rmtree(path)
+
+    bucket, prefix = parse_s3_uri(prefix_uri)
+    # Run Mountpoint in background mode, and arrange for it to unmount when this script exits
+    tempdir = tempfile.mkdtemp(prefix="s3dataset_")
+    binary = mountpoint_path or "mount-s3"
+    args = additional_args or []
+    subprocess.run([binary, bucket, tempdir] + args, check=True)
+    atexit.register(teardown, tempdir)
+
+    # Now we can just read our dataset as if it were a local directory
+    return str(Path(tempdir) / prefix)
+
+
 def make_dataset(
     kind: str,
     sharding: Optional[DatasetSharding],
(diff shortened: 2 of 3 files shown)


s3torchbenchmarking/conf/dataloading.yaml
```diff
@@ -9,5 +9,5 @@ hydra:
   mode: MULTIRUN
   sweeper:
     params:
-      dataloader: s3iterabledataset, fsspec
+      dataloader: s3iterabledataset, fsspec, mountpoint
```
muddyfish, line 12: This is the default right? Do we want to have everyone running benchmarking to need MP installed?

s3torchbenchmarking/src/s3torchbenchmarking/benchmark.py
```diff
@@ -71,6 +74,39 @@ def from_conf(dataset_config: DictConfig):
             return DatasetSharding[dataset_config.sharding]
 
 
+def parse_s3_url(url: str) -> (str, str):
```
muddyfish, line 77: This is implemented in s3torchconnector. Why not import it? (I think a private import here is probably OK as it's the same repo)

s3torchbenchmarking/src/s3torchbenchmarking/benchmark.py
```diff
@@ -71,6 +74,39 @@ def from_conf(dataset_config: DictConfig):
...
+        subprocess.run(["sudo", "umount", path])
+        shutil.rmtree(path)
+
+    bucket, prefix = parse_s3_url(prefix_uri)
+    # Run Mountpoint in background mode, and arrange for it to unmount when this script exits
+    tempdir = tempfile.mkdtemp(prefix="s3dataset_")
+    binary = mountpoint_path or "mount-s3"
+    args = additional_args.split(" ") if additional_args else []
+    subprocess.run([binary, bucket, tempdir] + args, check=True)
+    atexit.register(teardown, tempdir)
```
muddyfish, line 104: Do we want to do this before starting MP?

s3torchbenchmarking/src/s3torchbenchmarking/benchmark.py
```diff
@@ -96,7 +132,9 @@ def make_dataset(
                 prefix_uri, region=region, transform=load_sample
             )
             return dataset
-    elif kind == "fsspec":
+    elif kind == "fsspec" or kind == "mountpoint":
```
muddyfish, line 135: `kind in ("fsspec", "mountpoint")`?

s3torchbenchmarking/src/s3torchbenchmarking/benchmark.py
vishalg0wda: Eager teardown registration sounds reasonable if we're anticipating some surface area of error between the mounting and registration - in this case they are executed one after the other - if the subprocess cmd failed, we shouldn't need to teardown anything.

s3torchbenchmarking/conf/dataloading.yaml
vishalg0wda: Agreed. This should be opt-in rather than enforced.

s3torchbenchmarking/src/s3torchbenchmarking/benchmark.py
```diff
@@ -96,7 +121,9 @@ def make_dataset(
                 prefix_uri, region=region, transform=load_sample
             )
             return dataset
-    elif kind == "fsspec":
+    elif kind in ("fsspec", "mountpoint"):
```
IsaevIlya, line 124: Should we use [FileLister](https://pytorch.org/data/main/generated/torchdata.datapipes.iter.FileLister.html#torchdata.datapipes.iter.FileLister)/[FileOpener](https://pytorch.org/data/main/generated/torchdata.datapipes.iter.FileOpener.html#torchdata.datapipes.iter.FileOpener) for `mountpoint` or there are no difference in performance with `FSSpecFileLister`/`FSSpecFileOpener` when dealing with files stored localy?

s3torchbenchmarking/src/s3torchbenchmarking/benchmark.py
```diff
@@ -71,6 +75,27 @@ def from_conf(dataset_config: DictConfig):
...
+) -> str:
+    def teardown(path: str):
+        subprocess.run(["sudo", "umount", path])
+        shutil.rmtree(path)
+
+    bucket, prefix = parse_s3_uri(prefix_uri)
+    # Run Mountpoint in background mode, and arrange for it to unmount when this script exits
+    tempdir = tempfile.mkdtemp(prefix="s3dataset_")
+    binary = mountpoint_path or "mount-s3"
+    args = additional_args.split(" ") if additional_args else []
```
dnanuti, line 91: Perhaps make additional_args `List[str] = None` and avoid the split?

s3torchbenchmarking/src/s3torchbenchmarking/benchmark.py
```diff
@@ -71,6 +75,27 @@ def from_conf(dataset_config: DictConfig):
             return DatasetSharding[dataset_config.sharding]
 
 
+def make_mountpoint(
+    prefix_uri: str,
+    mountpoint_path: Optional[str] = None,
+    additional_args: Optional[str] = None,
+) -> str:
+    def teardown(path: str):
+        subprocess.run(["sudo", "umount", path])
```
dnanuti, line 84: Should we add to the instructions that we need sudo?

s3torchbenchmarking/src/s3torchbenchmarking/benchmark.py
```diff
@@ -71,6 +75,27 @@ def from_conf(dataset_config: DictConfig):
...
+    binary = mountpoint_path or "mount-s3"
+    args = additional_args.split(" ") if additional_args else []
+    subprocess.run([binary, bucket, tempdir] + args, check=True)
+    atexit.register(teardown, tempdir)
+
+    # Now we can just read our dataset as if it were a local directory
+    return str(Path(tempdir) / prefix)
+
+
 def make_dataset(
```
dnanuti, line 99: Non-blocking: We need to simply this factory method.

vishalg0wda: Sure. What simplification did you have in mind?

s3torchbenchmarking/README.md
```diff
@@ -75,6 +75,17 @@ Then from this directory, install the dependencies:
...
+**(Optional) Install Mountpoint**
+
+Required only if you're running benchmarks using PyTorch with Mountpoint for S3.
+
+    wget https://s3.amazonaws.com/mountpoint-s3-release/latest/x86_64/mount-s3.rpm
+    sudo yum install ./mount-s3.rpm # For an RHEL system
+
+For other distros see [Installing Mountpoint for Amazon S3](https://github.com/awslabs/mountpoint-s3/blob/main/doc/INSTALL.md).
+
+_Note: Mountpoint benchmarks are currently only supported on *nix-based systems and relies on `sudo` capabilities._
```
dnanuti, line 87: nit: I think it's rely, but @muddyfish is the expert

s3torchbenchmarking/src/s3torchbenchmarking/benchmark.py
dnanuti: non-blocking: At the first glance, extract separately the logic for each kind, so that here we have just a method call - i.e.
```
if kind == "s3iterabledataset":
    create_iterable_dataset(...) 
```
As `make_dataset` will grow with the addition of supported dataset types, but the logic for each individual type can be separated.

dnanuti (approved):

s3torchbenchmarking/src/s3torchbenchmarking/benchmark.py
vishalg0wda: This is a good question. I haven't verified the difference so far. We should analyze this separately and pick the more performant implementation to make a fairer analysis(rather than expose this as another knob for users).

dnanuti (approved):

PR #161, opened 2024-02-29
Refactor UserAgent setup for extensibility

dnanuti:
## Description

Refactor User-Agent setup with an extensible configuration according to [RFC9110](https://www.rfc-editor.org/rfc/rfc9110#name-user-agent).

## Related items

- Enhancements for previously added https://github.com/awslabs/s3-connector-for-pytorch/pull/65

## Testing

- Unit tests
--------

### DIFF
--- a/s3torchconnector/src/s3torchconnector/_s3client/_s3client.py
+++ b/s3torchconnector/src/s3torchconnector/_s3client/_s3client.py
@@ -7,7 +7,6 @@ from functools import partial
 from typing import Optional, Any
 
 from s3torchconnector import S3Reader, S3Writer
-from s3torchconnector._version import user_agent_prefix
 
 from s3torchconnectorclient._mountpoint_s3_client import (
     MountpointS3Client,
@@ -16,6 +15,7 @@ from s3torchconnectorclient._mountpoint_s3_client import (
     GetObjectStream,
 )
 
+from s3torchconnector._user_agent import UserAgent
 
 """
 _s3client.py
@@ -32,11 +32,13 @@ def _identity(obj: Any) -> Any:
 
 
 class S3Client:
-    def __init__(self, region: str, endpoint: str = None):
+    def __init__(self, region: str, endpoint: str = None, user_agent: UserAgent = None):
         self._region = region
         self._endpoint = endpoint
         self._real_client = None
         self._client_pid = None
+        user_agent = user_agent or UserAgent()
+        self._user_agent_prefix = user_agent.prefix
 
     @property
     def _client(self) -> MountpointS3Client:
@@ -50,11 +52,15 @@ class S3Client:
     def region(self) -> str:
         return self._region
 
+    @property
+    def user_agent_prefix(self) -> str:
+        return self._user_agent_prefix
+
     def _client_builder(self) -> MountpointS3Client:
         return MountpointS3Client(
             region=self._region,
             endpoint=self._endpoint,
-            user_agent_prefix=user_agent_prefix,
+            user_agent_prefix=self._user_agent_prefix,
         )
 
     def get_object(
--- a/s3torchconnector/src/s3torchconnector/_user_agent.py
+++ b/s3torchconnector/src/s3torchconnector/_user_agent.py
@@ -9,6 +9,8 @@ from ._version import __version__
 
 class UserAgent:
     def __init__(self, comments: List[str] = None):
+        if comments and not isinstance(comments, list):
+            raise ValueError("Argument comments must be a List[str]")
         self._user_agent_prefix = f"{__package__}/{__version__}"
         self._comments = comments or []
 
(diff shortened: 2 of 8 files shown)


s3torchconnector/tst/unit/test_s3_client.py
```diff
@@ -48,3 +50,18 @@ def test_list_objects_log(s3_client: S3Client, caplog):
     with caplog.at_level(logging.DEBUG):
         s3_client.list_objects(TEST_BUCKET, TEST_KEY)
     assert f"ListObjects {S3_URI}" in caplog.messages
+
+
+def test_s3_client_default_user_agent():
+    s3_client = S3Client(region=TEST_REGION)
```
muddyfish, line 56: There's no test validating the user agent is correctly passed to out Mountpoint client

s3torchconnector/src/s3torchconnector/_s3client/_s3client.py
```diff
@@ -32,11 +32,12 @@ def _identity(obj: Any) -> Any:
 
 
 class S3Client:
-    def __init__(self, region: str, endpoint: str = None):
+    def __init__(self, region: str, endpoint: str = None, user_agent: UserAgent = None):
         self._region = region
         self._endpoint = endpoint
         self._real_client = None
         self._client_pid = None
+        self._user_agent = user_agent or UserAgent()
```
muddyfish, line 40: Are we expecting to mutate the user agent after it's been passed to an S3 client? If not this area can be simplified a lot

s3torchconnector/src/s3torchconnector/_user_agent.py
```diff
@@ -0,0 +1,24 @@
...
+        self._user_agent_prefix = f"{__package__}/{__version__}"
+        self._comments = comments or []
+        self._comments_separator = "; "
+
+    def add_comment(self, comment):
+        self._comments.append(comment)
+
+    @property
+    def prefix(self):
+        comments_str = f"{self._comments_separator.join(self._comments)}" or ""
```
muddyfish, line 21: No need for the or block here, or the f-string. I'm also not convinced we need the `_comments_separator` variable at all

s3torchconnector/src/s3torchconnector/_user_agent.py
```diff
@@ -0,0 +1,24 @@
...
+        if comments_str != "":
```
muddyfish, line 22: Nit: `if comments_str:` works here

s3torchconnector/tst/unit/test_user_agent.py
```diff
@@ -0,0 +1,33 @@
...
+    user_agent = UserAgent([""])
+    assert user_agent.prefix == f"s3torchconnector/{__version__}"
+
+
+def test_user_agent_creation_with_none_comments():
+    user_agent = UserAgent(None)
+    assert user_agent.prefix == f"s3torchconnector/{__version__}"
+
+
+def test_user_agent_creation_with_comments():
```
muddyfish, line 28: Not blocking: Might want to consider a hypothesis test that asserts the prefix always starts with `s3torchconnector/{__version__}`, and always includes the comments given in the string

s3torchconnector/tst/unit/test_s3_client.py
```diff
@@ -48,3 +52,29 @@ def test_list_objects_log(s3_client: S3Client, caplog):
...
+    )
+    expected_user_agent = (
+        f"s3torchconnector/{__version__} (component/version; metadata)"
+    )
+    assert s3_client.user_agent_prefix == expected_user_agent
+    assert s3_client._client.user_agent_prefix == expected_user_agent
+
+
+@given(text())
+def test_user_agent_always_starts_with_package_version(comments):
```
muddyfish, line 76: I don't think this is doing what you think it's doing :)
Perhaps `UserAgent` should assert it's given a list, not a string...

s3torchconnector/tst/unit/test_s3_client.py
```diff
@@ -48,3 +52,29 @@ def test_list_objects_log(s3_client: S3Client, caplog):
...
+    s3_client = S3Client(region=TEST_REGION, user_agent=UserAgent(comments))
+    expected_prefix = f"s3torchconnector/{__version__}"
+    assert s3_client.user_agent_prefix.startswith(expected_prefix)
+    assert s3_client._client.user_agent_prefix.startswith(expected_prefix)
```
muddyfish, line 80: We should also test that the comments we generate are present in the prefix

IsaevIlya (dismissed):

s3torchconnector/src/s3torchconnector/_user_agent.py
```diff
@@ -0,0 +1,22 @@
...
+from typing import List
+
+from ._version import __version__
+
+# https://www.rfc-editor.org/rfc/rfc9110#name-user-agent
+
+
+class UserAgent:
+    def __init__(self, comments: List[str] = None):
+        if comments and not isinstance(comments, list):
```
muddyfish, line 12: Nit: `if comments is not None and ...`

muddyfish (approved):

muddyfish (approved):

PR #168, opened 2024-03-05
Add `type: ignore` annotations in various places to make mypy pass

muddyfish:
## Description

'Fixes' `mypy` by adding `type: ignore` annotations to a few lines

## Additional context

No breaking changes

## Testing

`mypy s3torchconnector/src && mypy s3torchconnectorclient/python/src` passes

--------

### DIFF
--- a/.github/workflows/python-checks.yml
+++ b/.github/workflows/python-checks.yml
@@ -97,7 +97,7 @@ jobs:
           python -m pip install flake8 black mypy
           python -m pip install torch --extra-index-url https://download.pytorch.org/whl/cpu
           python -m pip install ./s3torchconnectorclient
-          python -m pip install ./s3torchconnector
+          python -m pip install ./s3torchconnector[lightning]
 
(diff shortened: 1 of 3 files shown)


dnanuti (approved):

.github/workflows/python-checks.yml
```diff
@@ -97,7 +97,7 @@ jobs:
           python -m pip install flake8 black mypy
           python -m pip install torch --extra-index-url https://download.pytorch.org/whl/cpu
           python -m pip install ./s3torchconnectorclient
-          python -m pip install ./s3torchconnector
+          python -m pip install ./s3torchconnector[lightning]
```
dnanuti, line 100: Why are we doing this? Shouldn't we install lightning only when running those tests?

dnanuti (changes requested):
We need to test that the library is working without lightning too, we shouldn't install it from the start.

dnanuti (approved):

PR #171, opened 2024-03-06
Provide information on how S3MapDataset works

IsaevIlya:
## Description
Add information about how S3MapDataset is able to support random access pattern. Explain that S3MapDataset has to build the list of all objects to support random access. This process could be resource demanding, so it would start only when user try to access the actual data. 

--------

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,16 +52,25 @@ End to end example of how to use `s3torchconnector` can be found under the
 
 The simplest way to use the S3 Connector for PyTorch is to construct a dataset, either a map-style or iterable-style 
 dataset, by specifying an S3 URI (a bucket and optional prefix) and the region the bucket is located in:
-```shell
+```py
 from s3torchconnector import S3MapDataset, S3IterableDataset
 
 # You need to update <BUCKET> and <PREFIX>
 DATASET_URI="s3://<BUCKET>/<PREFIX>"
 REGION = "us-east-1"
 
-map_dataset = S3MapDataset.from_prefix(DATASET_URI, region=REGION)
-
 iterable_dataset = S3IterableDataset.from_prefix(DATASET_URI, region=REGION)
+
+# Datasets are also iterators. 
+for object in iterable_dataset:
+  print(object.key)
+
+# S3MapDataset eagerly lists all the objects under the given prefix 
+# to provide support of random access.  
+# S3MapDataset build list of all objects at the first access to its elements or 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,17 +52,26 @@ End to end example of how to use `s3torchconnector` can be found under the
 
 The simplest way to use the S3 Connector for PyTorch is to construct a dataset, either a map-style or iterable-style 
 dataset, by specifying an S3 URI (a bucket and optional prefix) and the region the bucket is located in:
-```shell
+```py
```
dnanuti, line 55: The IDE suggests `python` when trying to update this section.

README.md
```diff
@@ -52,17 +52,26 @@ End to end example of how to use `s3torchconnector` can be found under the
...
 iterable_dataset = S3IterableDataset.from_prefix(DATASET_URI, region=REGION)
 
+# Datasets are also iterators. 
+for object in iterable_dataset:
+  print(object.key)
+
+# S3MapDataset eagerly lists all the objects under the given prefix 
+# to provide support of random access.  
+# S3MapDataset build list of all objects at the first access to its elements or 
+# at the first call to get count of its elements, whichever happens first.
```
dnanuti, line 71: Should we mention directly `len` here?

README.md
IsaevIlya: I have tried to turn on rich view, it works with just `py`

README.md
IsaevIlya: I want to have first part and second part of sentence to be written in the same manner. To be both descriptive or both of them using code.

README.md
```diff
@@ -52,17 +52,26 @@ End to end example of how to use `s3torchconnector` can be found under the
...
 # You need to update <BUCKET> and <PREFIX>
 DATASET_URI="s3://<BUCKET>/<PREFIX>"
 REGION = "us-east-1"
 
-map_dataset = S3MapDataset.from_prefix(DATASET_URI, region=REGION)
-
 iterable_dataset = S3IterableDataset.from_prefix(DATASET_URI, region=REGION)
 
+# Datasets are also iterators. 
+for object in iterable_dataset:
```
muddyfish, line 65: In this example, can we use `obj`, as `object` is a builtin and shadowing it isn't a good practice

muddyfish (approved):

dnanuti (approved):

Issue #166, opened 2024-03-05
An error occurs when launching the benchmark.


shiwk:
`RuntimeError: Input type (torch.cuda.FloatTensor) and weight type (torch.FloatTensor) should be the same`

The issue arises when running the benchmark code for the Vision Transformer (VIT). I believe it occurs because the input tensor is located on the CPU, while the model is on the GPU.

It will work with the addition of a single line of code:
`model = self.model.to(self.device)`

IsaevIlya:
Hello @shiwk,
Thank you for letting us know about the issue. We will have a look on it. Meanwhile, could you share the instance type on which you faced that issue?

shiwk:
> Hello @shiwk, Thank you for letting us know about the issue. We will have a look on it. Meanwhile, could you share the instance type on which you faced that issue?

Hi @IsaevIlya , I run it on my own server with a GPU, not on any service of AWS

IsaevIlya:
@shiwk, Thank you for bringing up this issue. I've successfully replicated it, and the resolution should be addressed by the  PR https://github.com/awslabs/s3-connector-for-pytorch/pull/172

PR #172, opened 2024-03-07
Move model's object to a correct device

IsaevIlya:
## Description
There are discrepancy in where training data and model are located. When [train_batch](https://github.com/awslabs/s3-connector-for-pytorch/blob/main/s3torchbenchmarking/src/s3torchbenchmarking/models.py#L139) function of  `ViT` class is called, the input parameters such as `data` and `target` are moved to available device, when `model` is stayed on the default device. To prevent such discrepancy the `model` is moved to the same device as `data` and `target`.
That problem was reported in https://github.com/awslabs/s3-connector-for-pytorch/issues/166 issue.

## Testing
Manually tested on instances with and without GPU (g5.8xlarge and c6a.4xlarge)

--------


muddyfish (approved):

dnanuti (approved):

vishalg0wda (commented):
Thanks for fixing this!

PR #177, opened 2024-03-12
Tell pytest to use current/root folder as working directory

IsaevIlya:
## Description
Pytest 8.1.1 starts to ignore `conftest.py` located in the root directory. That leads to error with loading `ci` profile for  hypothesis. To prevent that issue, we will tell pytest to use current/root folder as working folder.

## Testing
Tested changes locally 

--------


dnanuti:
This is breaking the CI for main as well, so let's target main for these changes.

PR #191, opened 2024-03-21
Expose S3ClientConfig

dnanuti:
## Description
We expose the following configuration flags with performance impact: `throughput_target_gbps` and `part_size`.

## Additional context

 - `throughput_target_gbps(float)`: Throughput target in Gigabits per second (Gbps) that we are trying to reach.
        10.0 Gbps by default (may change in future).
 - `part_size(int)`: Size, in bytes, of parts that files will be downloaded or uploaded in.
        Note: for saving checkpoints, the inner client will adjust the part size to meet the service limits.
        (max number of parts per upload is 10,000, minimum upload part size is 5 MiB).
        Part size must have values between 5MiB and 5GiB.
        8MB by default (may change in future).

## Testing

- Unit tests

--------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+### New features
+* Expose `throughput_target_gbps` and `part_size` configurations of the inner S3 client.
+
 ## v1.2.1 (March 14, 2024)
 
 ### Breaking changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Unreleased
 
 ### New features
-* Expose `throughput_target_gbps` and `part_size` configurations of the inner S3 client.
+* Expose a new class, S3ClientConfig, with `throughput_target_gbps` and `part_size` parameters of the inner S3 client.
 
 ## v1.2.1 (March 14, 2024)
 
--- a/s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
+++ b/s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
@@ -6,11 +6,9 @@ from dataclasses import dataclass
 @dataclass(frozen=True)
 class S3ClientConfig:
     """A dataclass exposing configurable parameters for the S3 client.
-    Returns a config wrapper object.
 
     Args:
     throughput_target_gbps(float): Throughput target in Gigabits per second (Gbps) that we are trying to reach.
-        You can also use get_recommended_throughput_target_gbps() to get recommended value for your system.
         10.0 Gbps by default (may change in future).
     part_size(int): Size, in bytes, of parts that files will be downloaded or uploaded in.
         Note: for saving checkpoints, the inner client will adjust the part size to meet the service limits.
--- a/s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
+++ b/s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
@@ -1,0 +1,23 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  // SPDX-License-Identifier: BSD
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class S3ClientConfig:
+    """A dataclass exposing configurable parameters for the S3 client.
+    Returns a config wrapper object.
+
+    Args:
+    throughput_target_gbps(float): Throughput target in Gigabits per second (Gbps) that we are trying to reach.
+        You can also use get_recommended_throughput_target_gbps() to get recommended value for your system.
+        10.0 Gbps by default (may change in future).
+    part_size(int): Size, in bytes, of parts that files will be downloaded or uploaded in.
+        Note: for saving checkpoints, the inner client will adjust the part size to meet the service limits.
+        (max number of parts per upload is 10,000, minimum upload part size is 5 MiB).
+        Part size must have values between 5MiB and 5GiB.
+        8MB by default (may change in future).
+    """
+
+    throughput_target_gbps: float = 10.0
+    part_size: int = 8 * 1024 * 1024
--- a/s3torchconnector/tst/unit/test_s3_client.py
+++ b/s3torchconnector/tst/unit/test_s3_client.py
@@ -4,12 +4,14 @@ import logging
 import pytest
 
 from hypothesis import given
-from hypothesis.strategies import lists, text
+from hypothesis.strategies import lists, text, integers, floats
 from unittest.mock import MagicMock
+
+from s3torchconnectorclient._mountpoint_s3_client import S3Exception
 
 from s3torchconnector._user_agent import UserAgent
 from s3torchconnector._version import __version__
-from s3torchconnector._s3client import S3Client, MockS3Client
+from s3torchconnector._s3client import S3Client, MockS3Client, S3ClientConfig
 
 TEST_BUCKET = "test-bucket"
 TEST_KEY = "test-key"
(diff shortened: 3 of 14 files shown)


s3torchconnectorclient/python/src/s3torchconnectorclient/s3_connector_client_config.py
```diff
@@ -0,0 +1,10 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  // SPDX-License-Identifier: BSD
+
+
+class S3ConnectorClientConfig:
```
muddyfish, line 5: NB: This can be a dataclass

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## Unreleased
+
+### New features
+* Expose `throughput_target_gbps` and `part_size` configurations of the inner S3 client.
```
muddyfish, line 4: Not blocking: Perhaps mention this is exposed through a new `S3ClientConfig` class/argument?

s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
```diff
@@ -0,0 +1,23 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  // SPDX-License-Identifier: BSD
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class S3ClientConfig:
+    """A dataclass exposing configurable parameters for the S3 client.
+    Returns a config wrapper object.
```
muddyfish, line 9: A config wrapper object? I'd probably remove this line as it should be fairly obvious that instantiating a class gives you an instance of that class

s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
```diff
@@ -0,0 +1,23 @@
...
+
+    Args:
+    throughput_target_gbps(float): Throughput target in Gigabits per second (Gbps) that we are trying to reach.
+        You can also use get_recommended_throughput_target_gbps() to get recommended value for your system.
```
muddyfish, line 13: `get_recommended_throughput_target_gbps` doesn't exist

s3torchconnector/tst/unit/test_s3_client.py
```diff
@@ -82,3 +84,34 @@ def test_user_agent_always_starts_with_package_version(comments):
     if comments_str:
         assert comments_str in s3_client.user_agent_prefix
         assert comments_str in s3_client._client.user_agent_prefix
+
+
+@given(
+    part_size=integers(min_value=5 * 1024, max_value=5 * 1024 * 1024),
```
muddyfish, line 90: Not blocking: Might be worth giving explicit examples for min and max as in https://hypothesis.readthedocs.io/en/latest/reproducing.html#hypothesis.example to ensure they're always tested

s3torchconnector/tst/unit/test_s3_client.py
```diff
@@ -82,3 +84,34 @@ def test_user_agent_always_starts_with_package_version(comments):
...
+    throughput_target_gbps=floats(min_value=10.0, max_value=100.0),
+)
+def test_s3_client_custom_config(part_size: int, throughput_target_gbps: float):
+    # Part size must have values between 5MiB and 5GiB
+    part_size = part_size * 1024
```
muddyfish, line 95: Not blocking: If it doesn't work to give a part size that's not a multiple of 1kb, we should add a test for that, otherwise we should do the multiplication in the given

s3torchconnector/tst/unit/test_s3_client.py
```diff
@@ -82,3 +84,34 @@ def test_user_agent_always_starts_with_package_version(comments):
...
+    s3_client = S3Client(
+        region=TEST_REGION,
+        s3client_config=S3ClientConfig(
+            part_size=part_size,
+            throughput_target_gbps=throughput_target_gbps,
+        ),
+    )
+    assert s3_client._client.part_size == part_size
+    assert abs(s3_client._client.throughput_target_gbps - throughput_target_gbps) < 1e-9
```
muddyfish, line 104: Comment for all of these, why don't we do an equality assertion here?
If that doesn't work, pytest actually has a helper for this: https://stackoverflow.com/a/39623614

s3torchconnector/tst/unit/test_s3_client.py
```diff
@@ -82,3 +84,34 @@ def test_user_agent_always_starts_with_package_version(comments):
...
+
+
+def test_s3_client_invalid_part_size_config():
+    with pytest.raises(
+        S3Exception,
+        match="invalid configuration: part size must be at between 5MiB and 5GiB",
```
muddyfish, line 110: Nit: No test for part size over 5GiB

s3torchconnector/tst/unit/test_s3_client_config.py
```diff
@@ -0,0 +1,42 @@
...
+#  // SPDX-License-Identifier: BSD
+from hypothesis import given
+from hypothesis.strategies import integers, floats
+
+from s3torchconnector import S3ClientConfig
+
+
+def test_default():
+    config = S3ClientConfig()
+    assert config is not None
```
muddyfish, line 11: Not blocking: The is not None check isn't needed. It's definitely going to be an instance of `S3ClientConfig`. At the very most I'd do an isinstance check

s3torchconnector/tst/unit/test_s3_client.py
dnanuti: It should work, but I wanted to emphasize we're comparing floats

s3torchconnector/tst/unit/test_s3_client.py
```diff
@@ -82,3 +88,36 @@ def test_user_agent_always_starts_with_package_version(comments):
...
+        s3client_config=S3ClientConfig(
+            part_size=part_size,
+            throughput_target_gbps=throughput_target_gbps,
+        ),
+    )
+    assert s3_client._client.part_size == part_size
+    assert s3_client._client.throughput_target_gbps == throughput_target_gbps
+
+
+@pytest.mark.parametrize("part_size", [1, 2 * KiB, 6 * GiB])
```
muddyfish, line 112: Could we add `5MiB-1` and `5GiB+1` as tests?

s3torchconnector/tst/unit/test_s3_client.py
```diff
@@ -82,3 +88,36 @@ def test_user_agent_always_starts_with_package_version(comments):
...
+    with pytest.raises(
+        S3Exception,
+        match="invalid configuration: part size must be at between 5MiB and 5GiB",
+    ):
+        s3_client = S3Client(
+            region=TEST_REGION,
+            s3client_config=S3ClientConfig(part_size=part_size),
+        )
+        # The client is lazily initialized
+        assert s3_client._client.part_size is not None
```
muddyfish, line 123: Couldn't we do an == check here?

muddyfish (approved):

s3torchconnector/tst/unit/test_s3_client.py
dnanuti: It fails before getting to verify the assert, when is trying to initialize MP client with the invalid part_size value, that's why I avoided the == assertion.

s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
```diff
@@ -0,0 +1,21 @@
...
+    throughput_target_gbps(float): Throughput target in Gigabits per second (Gbps) that we are trying to reach.
+        10.0 Gbps by default (may change in future).
+    part_size(int): Size, in bytes, of parts that files will be downloaded or uploaded in.
+        Note: for saving checkpoints, the inner client will adjust the part size to meet the service limits.
+        (max number of parts per upload is 10,000, minimum upload part size is 5 MiB).
+        Part size must have values between 5MiB and 5GiB.
+        8MB by default (may change in future).
+    """
+
+    throughput_target_gbps: float = 10.0
```
IsaevIlya, line 20: NIT: It would be better to define default values as constants to re-use them in tests, with that tests will continue to pass even if we will change default values.

PR #197, opened 2024-04-02
Build macOS wheels and integration tests

jamesbornholt:
This is pretty simple, mostly just adding macos-13 (x86) and macos-14 (arm64) runners to the matrix. We also need to explicitly disable MPS for PyTorch to run correctly on the GHA runners which don't have access to that hardware.

--------
dnanuti@: Build on macos runners had occasionally a flaky behaviour, as `actions/setup-python` requires `/Users/runner/hostedtoolcache` to exist to work properly with macosx due to [fixed shared library path](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#macos).
Additionally, `pipx` was not always properly configured, depending on the runner image, encountered [this](https://github.com/actions/runner-images/issues/9607) on macos-14 arm64 runners too ([example run](https://github.com/awslabs/s3-connector-for-pytorch/actions/runs/8630946711/job/23658476975)).

Also worth mentioning:
>  Warning: While cibuildwheel can build CPython 3.8 universal2/arm64 wheels, we cannot test the arm64 part of them, even when running on an Apple Silicon machine. This is because we use the x86_64 installer of CPython 3.8. See the discussion in https://github.com/pypa/cibuildwheel/pull/1169 for the details. To silence this warning, set `CIBW_TEST_SKIP: "cp38-macosx_*:arm64"`.

This applies just for the build for 3.8 macosx arm64 and had the same behaviour with the [initial changes](https://github.com/awslabs/s3-connector-for-pytorch/commit/e353788fda75ed8f47091b87f9162d6f4863cd57).

--------

### DIFF
--- a/.github/workflows/python-integration.yml
+++ b/.github/workflows/python-integration.yml
@@ -19,11 +19,12 @@ env:
 jobs:
   integration-test:
     name: Integration tests
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.runner }}
     environment: ${{ inputs.environment }}
     strategy:
       fail-fast: false
       matrix:
+        runner: [ubuntu-22.04, macos-13]
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
         test-run:
           - name: "S3"
@@ -53,17 +54,13 @@ jobs:
           aws-region: ${{ vars.S3_REGION }}
 
       - name: Set up stable Rust
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          override: true
+        uses: dtolnay/rust-toolchain@stable
 
       - name: Restore Cargo cache
         id: restore-cargo-cache
         uses: actions/cache/restore@v3
         with:
           path: |
-            ~/.cargo/bin/
             ~/.cargo/registry/index/
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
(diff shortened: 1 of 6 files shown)


.github/workflows/python-integration.yml
```diff
@@ -53,17 +54,13 @@ jobs:
           aws-region: ${{ vars.S3_REGION }}
 
       - name: Set up stable Rust
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          override: true
+        uses: dtolnay/rust-toolchain@stable
```
dnanuti, line 57: Why have we changed this? 
Should we update the other workflows too: https://github.com/search?q=repo%3Aawslabs%2Fs3-connector-for-pytorch%20actions-rs%2Ftoolchain%40v1&type=code?

.github/workflows/wheels.yml
```diff
@@ -62,16 +71,29 @@ jobs:
...
+        if: ${{ matrix.builder.arch != 'aarch64' }}
+        uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version: "3.12"
 
       - name: Install pipx
         run: |
-          python -m pip install --upgrade pipx 
+          which python
```
muddyfish, line 92: Are these `which` lines still wanted?

dnanuti: I used them for debugging, but left them on purpose, as they helped to understand the problem, and I am expecting changes for pip setup in the upcoming runner updates.

Issue #199, opened 2024-04-04
Improve performance for loading checkpoints


IsaevIlya:
### Tell us more about this new feature.

Hello,
There is the difference in performance for storing and loading "big" checkpoints (~20Gb), while using s3-connector-for-pytorch and PyTorch. When saving and loading 17.4Gb checkpoint on g5.12xlarge instance in us-west-2 region the time needed for upload is ~17 seconds, but time for load is ~30 seconds. Is it possible to improve loading time?
Bellow is the script that was used for measuring time:

```py
import torch
import uuid
import boto3
import s3torchconnector
import time
from transformers import AutoModelForMaskedLM

print("Start")

# 270mb model
model = AutoModelForMaskedLM.from_pretrained("distilbert/distilbert-base-uncased")
print("Loaded model")

# Increase the embedding size
model.resize_token_embeddings(6000000)
print("Resized model")

checkpoint_bucket = f"checkpoint-test{uuid.uuid4()}"
checkpoint_file = "test.chkpt"
checkpoint_path = f"s3://{checkpoint_bucket}/{checkpoint_file}"
region = "us-west-2"

# create bucket in the same region, where you run script
s3_client = boto3.client('s3', region_name=region)
s3_client.create_bucket(Bucket=checkpoint_bucket, CreateBucketConfiguration={'LocationConstraint': region})
print(f"Bucket '{checkpoint_bucket}' created successfully.")

part_size = 8 * 1024 * 1024
throughput_target_gbps = 100

print(f"part size is {part_size} bytes")
print(f"throughput target is {throughput_target_gbps} gbps")

config = s3torchconnector.S3ClientConfig(part_size = part_size, throughput_target_gbps=throughput_target_gbps)
s3checkpoint = s3torchconnector.S3Checkpoint(region=region, s3client_config=config)
    
# Measure upload time
t1 = time.time()
with s3checkpoint.writer(checkpoint_path) as writer:
    torch.save(model.state_dict(), writer)
t2 = time.time()

# list files in bucket
response = s3_client.list_objects_v2(Bucket=checkpoint_bucket)
file_name = [obj['Key'] for obj in response['Contents']]
file_size = [obj['Size'] for obj in response['Contents']]
print(f"File in bucket '{checkpoint_bucket}': {file_name}, file size {file_size}")
print(f"upload time to {checkpoint_path} is  {t2 - t1}")

# load model from checkpoint
print(checkpoint_path)
loaded = None
[...]

PR #200, opened 2024-04-04
Add implementation of `readinto` to decrease amount of copy operations

IsaevIlya:
## Description
Changes to address https://github.com/awslabs/s3-connector-for-pytorch/issues/199
PyTorch compresses a model's data before storing it as a checkpoint. When loading a checkpoint with PyTorch, we need to decompress the stored compressed data. The process of loading this compressed data is not executed sequentially. During the loading process, it jumps back and forth through the stream representing the checkpoint to read separate chunks of data. This required us to create copies of these data parts and return them to PyTorch for processing. However, PyTorch supports an alternative way to obtain a copy of that data. If the object passed to `torch.load` supports the `readinto` method, then PyTorch will provide a pre-allocated buffer for writing that data directly, without the need for creating intermediate copies. See [here](https://github.com/pytorch/pytorch/blob/1e8d4b389b5f03cea191ed558051f036fe04f92d/torch/csrc/jit/python/init.cpp#L1500C11-L1511)

## Additional context
Add to `s3reader` implementation of `readinto` to decrease amount of copy operations. That allows to decrease the checkpoint loading time from 28 to 15 seconds on g5.12xlarge instance, while running test script from https://github.com/awslabs/s3-connector-for-pytorch/issues/199

## Related items
https://github.com/awslabs/s3-connector-for-pytorch/issues/199

## Testing
- Unit test was updated to test new `readinto` method of `s3reader`.
- Integration test for saving/loading checkpoint was extended to run test for ~40mb checkpoint, that will be bigger than default part_size.
- Test script from https://github.com/awslabs/s3-connector-for-pytorch/issues/199 was used to confirm performance improvement during checkpoint loading.

--------

### DIFF
--- a/s3torchconnector/src/s3torchconnector/s3reader.py
+++ b/s3torchconnector/src/s3torchconnector/s3reader.py
@@ -53,6 +53,34 @@ class S3Reader(io.BufferedIOBase):
         if self._stream is None:
             self._stream = self._get_stream()
 
+    def readinto(self, buf) -> int:
+        """Read up to len(buf) bytes into a pre-allocated, writable bytes-like object buf.
+        Return the number of bytes read. If no bytes available, zero is returned.
+
+        Args:
+            buf : writable bytes-like object
+
+        Returns:
+            int : numer of bytes read or zero, if no bytes available
+        """
+        buf_size = len(buf)
+        if self._position_at_end() or buf_size == 0:
+            # If no bytes are available or no place to write data, zero should be returned
+            return 0
+
+        self.prefetch()
+        assert self._stream is not None
+
+        cur_pos = self._position
+        # preload enough bytes in buffer
+        self.seek(buf_size, SEEK_CUR)
+        # restore position, before starting to write into buf
+        self._buffer.seek(cur_pos)
+        size = self._buffer.readinto(buf)
+        self._position = self._buffer.tell()
+
+        return size
+
     def read(self, size: Optional[int] = None) -> bytes:
         """Read up to size bytes from the object and return them.
 
@@ -82,7 +110,9 @@ class S3Reader(io.BufferedIOBase):
         if size is None or size < 0:
             # Special case read() all to use O(n) algorithm
             self._buffer.seek(0, SEEK_END)
-            self._buffer.write(b"".join(self._stream))
+            for batch in self._stream:
+                self._buffer.write(batch)
+
             # Once we've emptied the buffer, we'll always be at EOF!
             self._size = self._buffer.tell()
         else:
--- a/s3torchconnector/tst/unit/test_s3reader.py
+++ b/s3torchconnector/tst/unit/test_s3reader.py
@@ -21,7 +21,6 @@ logging.getLogger().setLevel(1)
 
 log = logging.getLogger(__name__)
 
-
 TEST_BUCKET = "test-bucket"
 TEST_KEY = "test-key"
 MOCK_OBJECT_INFO = Mock(ObjectInfo)
@@ -182,14 +181,17 @@ def test_over_read(stream: List[bytes], overread: int):
 def test_seeks_end():
     s3reader = S3Reader(TEST_BUCKET, TEST_KEY, lambda: None, lambda: iter([]))
     s3reader._size = 10
+    buf = memoryview(bytearray(10))
 
     assert s3reader.seek(0, SEEK_END) == 10
     assert s3reader.tell() == 10
     assert s3reader.read() == b""
+    assert s3reader.readinto(buf) == 0
 
     assert s3reader.seek(0, SEEK_CUR) == 10
     assert s3reader.tell() == 10
     assert s3reader.read() == b""
+    assert s3reader.readinto(buf) == 0
 
 
 def test_not_writable():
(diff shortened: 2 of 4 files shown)


s3torchconnector/tst/unit/test_s3reader.py
```diff
@@ -301,3 +303,110 @@ def test_s3reader_writes_size_after_read_all_explicit(stream: List[bytes]):
...
+        total_read = should_read_bytes_in_first_pass + should_read_bytes_in_second_pass
+        assert s3reader.tell() == total_read
+        # confirm that read data is the same as in source
+        assert (
+            buf[:should_read_bytes_in_second_pass]
+            == all_data[should_read_bytes_in_first_pass:total_read]
+        )
+        if total_length < total_read:
+            assert s3reader._size == total_read
+
```
muddyfish, line 371: Could we have a simple test that verifies `readinto` with a buf with length equal or greater than of the total object always reads the entire thing?

s3torchconnector/src/s3torchconnector/s3reader.py
```diff
@@ -53,6 +53,34 @@ def prefetch(self) -> None:
         if self._stream is None:
             self._stream = self._get_stream()
 
+    def readinto(self, buf) -> int:
+        """Read up to len(buf) bytes into a pre-allocated, writable bytes-like object buf.
+        Return the number of bytes read. If no bytes available, zero is returned.
```
muddyfish, line 58: Nit: "If no bytes *are* available, zero is returned"

muddyfish:
Happy to approve once the lints are done

s3torchconnector/tst/unit/test_s3reader.py
```diff
@@ -301,3 +303,106 @@ def test_s3reader_writes_size_after_read_all_explicit(stream: List[bytes]):
     assert s3reader.read(1) == b""
     # Once we've read past the end, we know how big the file is
     assert s3reader._size == total_length
+
+
+@given(
+    lists(binary(min_size=20, max_size=30), min_size=0, max_size=2),
+    integers(min_value=0, max_value=10),
+)
+def test_s3reader_readinto_buffer_smaller_then_chunks(
```
dnanuti, line 312: nit: than*

s3torchconnector/src/s3torchconnector/s3reader.py
```diff
@@ -53,6 +53,34 @@ def prefetch(self) -> None:
...
+
+        Returns:
+            int : numer of bytes read or zero, if no bytes available
+        """
+        buf_size = len(buf)
+        if self._position_at_end() or buf_size == 0:
+            # If no bytes are available or no place to write data, zero should be returned
+            return 0
+
+        self.prefetch()
```
dnanuti, line 71: Generally looks good to me. Only one thing that I can't get the grasp on. 
> readinto method, then PyTorch will provide a pre-allocated buffer for writing that data directly, without the need for creating intermediate copies

How large will this be? Can it be smaller than a part_size?

dnanuti: Also, might be a bit of a stretch, but can we also measure/assert memory usage? Since the purpose of the fix was to decrease the amount of copy operations, we should have a way to get a sense of that.

IsaevIlya: Yes, it is possible. The amounts of bytes read by PyTorch is not related to the part_size, it depends on the model structure and how it was compressed. Based on the patter I saw during testing, PyTorch will read header block that is less than 100 bytes and then read the block itself that could be bigger or smaller than part_size.

muddyfish (approved):

s3torchconnector/src/s3torchconnector/s3reader.py
dnanuti: Two more things:
- Do we know if the Lightning checkpoints behave the same? Will they leverage the `readinto` method?
- For testing the completeness of reads, I think we should add e2e tests where we pass a client config varying the part_size to `S3Checkpoint` (and `S3LightningCheckpoint` if that's impacted too).

dnanuti (approved):

s3torchconnector/src/s3torchconnector/s3reader.py
IsaevIlya: > Do we know if the Lightning checkpoints behave the same? Will they leverage the readinto method?

They should behave the same, because they implementation based on using torch.save/torch.load methods.

>     For testing the completeness of reads, I think we should add e2e tests where we pass a client config varying the part_size to S3Checkpoint (and S3LightningCheckpoint if that's impacted too).

The reading of data using the different part_size is covered by unit tests. I'm not sure that we need to also include that check as part of integration tests as it will increase the time for running e2e even more.

> Also, might be a bit of a stretch, but can we also measure/assert memory usage?

From the logical view, we eliminated one copy. I haven't run the profiler tests on memory, as main goal and main measure of success was time needed for downloading data.

> L.E. Otherwise put, what can we add to our tests to ensure we do not introduce regressions over time?

I think we need to work on that under separate tasks as it requires researching what limits to use to prevent fluky results or to relaxed limits.

PR #228, opened 2024-08-31
support tell for s3writer

ryxli:
## Description

Add tell to writer to get current position of putobjectstream

## Additional context

- [ x] I have updated the CHANGELOG or README if appropriate

## Testing

--------

### DIFF
--- a/s3torchconnector/tst/unit/test_s3writer.py
+++ b/s3torchconnector/tst/unit/test_s3writer.py
@@ -1,0 +1,55 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  // SPDX-License-Identifier: BSD
+
+from io import BytesIO
+from typing import List, Tuple
+from unittest.mock import Mock
+
+import pytest
+from hypothesis import given
+from hypothesis.strategies import lists, binary, composite
+from s3torchconnectorclient._mountpoint_s3_client import ObjectInfo, PutObjectStream
+
+from s3torchconnector import S3Writer
+
+MOCK_OBJECT_INFO = Mock(ObjectInfo)
+MOCK_STREAM = Mock(PutObjectStream)
+
+
+@composite
+def bytestream_and_lengths(draw):
+    byte_array = draw(lists(binary(min_size=1, max_size=5000)))
+    lengths = [len(b) for b in byte_array]
+    return byte_array, lengths
+
+
+def test_s3writer_creation():
+    s3writer = S3Writer(MOCK_STREAM)
+    assert s3writer
+    assert isinstance(s3writer.stream, PutObjectStream)
+
+
+@pytest.mark.parametrize(
+    "stream",
+    [
+        [b"1", b"2", b"3"],
+        [],
+        [b"hello!"],
+    ],
+)
+def test_s3writer_write(stream):
+    s3writer = S3Writer(MOCK_STREAM)
+    s3writer.write(stream)
+    s3writer.close()
+    MOCK_STREAM.write.assert_called_with(stream)
+
+
+@given(bytestream_and_lengths())
+def test_s3writer_tell(stream_and_lengths: Tuple[List[bytes], List[int]]):
+    with S3Writer(MOCK_STREAM) as s3writer, BytesIO() as bytewriter:
+        for data, length in zip(*stream_and_lengths):
+            b_length = s3writer.write(data)
+            bytewriter.write(data)
+
+            assert b_length == length
+            assert bytewriter.tell() == s3writer.tell()
--- a/s3torchconnector/src/s3torchconnector/s3writer.py
+++ b/s3torchconnector/src/s3torchconnector/s3writer.py
@@ -12,8 +12,10 @@ class S3Writer(io.BufferedIOBase):
 
     def __init__(self, stream: PutObjectStream):
         self.stream = stream
+        self._position = 0
 
     def __enter__(self):
+        self._position = 0
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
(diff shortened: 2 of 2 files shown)


IsaevIlya (changes requested):
Hello @ryxli,
Thank you for your contribution. Could you kindly:
- provide a brief reasoning for adding the `tell` method to the S3Writer class.
- include unit/integration tests to ensure the new functionality works as expected.

ryxli:
Hi @IsaevIlya 

> provide a brief reasoning for adding the tell method to the S3Writer class.

More generally, adding `tell` method to S3Writer class makes sense because it extends the `io.BufferedIOBase` and enhances the interoperability and usability of the S3Writer class as a in place drop in for a `Pathlike` object. Without it, parent class functionality like seek(), etc.. won't work, even if S3Writer doesn't support the Buffer protocol for now.
The `S3Reader` class implements the tell() method in a similar fashion: https://github.com/awslabs/s3-connector-for-pytorch/blob/main/s3torchconnector/src/s3torchconnector/s3reader.py#L200


More specifically, I can give an example of one use case where having the stream position is useful:

When integrating the MountPointS3Client with pytorch distributed checkpointing, one of the API exposed is [torch.distributed.checkpoint.FileSystemReader](https://pytorch.org/docs/stable/_modules/torch/distributed/checkpoint/filesystem.html#FileSystemWriter)
For each TensorObject / ByteObject in the sharded state dict, the torch implementation will create a plan to write each object into a stream: 
```
            with create_stream(file_name, "wb") as stream:
                for write_item in bytes_w:
                    data = planner.resolve_data(write_item)
                    write_results.append(
                        _write_item(stream, data, write_item, storage_key)
                    )

                for tensor, write_item in loader.values():
                    assert tensor.is_cpu
                    write_results.append(
                        _write_item(stream, tensor, write_item, storage_key)
                    )
```

The `_write_item` helper is written as follows:
```
def _write_item(
    stream: io.IOBase,
    data: Union[io.BytesIO, torch.Tensor],
    write_item: WriteItem,
    storage_key: str,
) -> WriteResult:
    offset = stream.tell()

    if write_item.type == WriteItemType.BYTE_IO:
        assert isinstance(data, io.BytesIO)
        stream.write(data.getbuffer())
    else:
        assert isinstance(data, torch.Tensor)
        assert data.device == torch.device("cpu")
        torch.save(data, cast(IO[bytes], stream))
    length = stream.tell() - offset

    return WriteResult(
        index=write_item.index,
        size_in_bytes=length,
        storage_data=_StorageInfo(storage_key, offset, length),
    )
```
There is an intermediate API which stores write results, and is used later when the distributed checkpoint plan is executed in a queue like manner.


With the current state of S3Writer, it is possible to get the position of the stream via `ret_val` of `S3Writer.write`, but it would require additional modification to the Pytorch source code, rather than a simple extension of `FileSystemWriter` via drop in replacement. Something similar to:

```
def _modified_write_item(
    stream: S3Writer,
    data: Union[io.BytesIO, torch.Tensor],
    write_item: WriteItem,
    storage_key: str,
) -> WriteResult:
    offset = 0 # some start of position, with S3Writer, on _enter_, this should always be zero since we write to s3 object from scratch rather than appending to existing s3 object.

    if write_item.type == WriteItemType.BYTE_IO:
        assert isinstance(data, io.BytesIO)
        length = stream.write(data.getbuffer()) # <--- get length from write return here
    else:
        assert isinstance(data, torch.Tensor)
        assert data.device == torch.device("cpu")
        torch.save(data, cast(IO[bytes], stream))  # <--- would need additional modification to calculate size of tensor since torch.save does not return length, and there is no way to retrieve the length of the data written using S3Writer when passed into torch.save
        # length = something

    return WriteResult(
        index=write_item.index,
        size_in_bytes=length,
        storage_data=_StorageInfo(storage_key, offset, length),
    )

```


> include unit/integration tests to ensure the new functionality works as expected.

Sure, I can add. I didn't include because I couldn't find any existing unit tests for S3Writer. S3Reader I found here: https://github.com/awslabs/s3-connector-for-pytorch/blob/main/s3torchconnector/tst/unit/test_s3reader.py

ryxli:
added tests

s3torchconnector/tst/unit/test_s3writer.py
```diff
@@ -0,0 +1,65 @@
...
+
+logging.basicConfig(
+    format="%(levelname)s %(name)s %(asctime)-15s %(filename)s:%(lineno)d %(message)s"
+)
+logging.getLogger().setLevel(1)
+
+log = logging.getLogger(__name__)
+
+TEST_BUCKET = "test-bucket"
+TEST_KEY = "test-key"
```
IsaevIlya, line 24: Nit: Could you please remove `TEST_BUCKET` and `TEST_KEY` as their are not used anywhere, other then that I'm happy to approve.

ryxli:
removed unused code.

tests passed:

```
% pytest s3torchconnector/tst/unit/test_s3writer.py -s
============================================================= test session starts =============================================================
platform linux -- Python 3.10.14, pytest-8.3.2, pluggy-1.5.0
rootdir: /fsx-Training/shopqa-training-fsx-prod-us-east-1/home/rynli/workspace/s3-connector-for-pytorch/s3torchconnector
configfile: pyproject.toml
plugins: hypothesis-6.111.2, timeout-2.3.1
collected 5 items                                                                                                                             

s3torchconnector/tst/unit/test_s3writer.py .....

============================================================== warnings summary ===============================================================
../../../../../../opt/conda/envs/s3connector/lib/python3.10/site-packages/torch/_subclasses/functional_tensor.py:258
  /opt/conda/envs/s3connector/lib/python3.10/site-packages/torch/_subclasses/functional_tensor.py:258: UserWarning: Failed to initialize NumPy: No module named 'numpy' (Triggered internally at ../torch/csrc/utils/tensor_numpy.cpp:84.)
    cpu = _conversion_method_template(device=torch.device("cpu"))

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
======================================================== 5 passed, 1 warning in 3.69s =========================================================
```

IsaevIlya (approved):
Thank you for your contribution, LGTM

PR #238, opened 2024-10-07
feat: add support for S3 `copy` API

matthieu-d4r:
Add support for S3 `copy` object API
(https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html).


s3torchconnectorclient/python/tst/integration/test_mountpoint_s3_integration.py
```diff
@@ -404,6 +404,44 @@ def test_delete_object_invalid_bucket(
...
+    # set up
+    client.delete_object(sample_directory.bucket, dst_key)
+
+    client.copy_object(src_key, dst_key)
+
+    src_obj = client.get_object(sample_directory.bucket, src_key)
+    dst_obj = client.get_object(sample_directory.bucket, dst_key)
+
+    assert dst_obj.key == dst_key
+    assert str(dst_obj) == str(src_obj)
```
muddyfish, line 423: Is this meant to compare the object contents? I don't think it would do that.

s3torchconnectorclient/python/tst/integration/test_mountpoint_s3_integration.py
```diff
@@ -404,6 +404,44 @@ def test_delete_object_invalid_bucket(
...
+
+    # tear down
+    client.delete_object(
```
muddyfish, line 426: This should probably be in a `finally` clause

s3torchconnectorclient/python/src/s3torchconnectorclient/_mountpoint_s3_client.pyi
```diff
@@ -35,6 +35,7 @@ class MountpointS3Client:
     ) -> ListObjectStream: ...
     def head_object(self, bucket: str, key: str) -> ObjectInfo: ...
     def delete_object(self, bucket: str, key: str) -> None: ...
+    def copy_object(self, src_bucket: str, src_key: str, dst_bucket: str, dst_prefix: str) -> ObjectInfo: ...
```
muddyfish, line 38: The rust implementation doesn't return an `ObjectInfo`

s3torchconnectorclient/python/tst/integration/test_mountpoint_s3_integration.py
```diff
@@ -404,6 +404,44 @@ def test_delete_object_invalid_bucket(
         )
 
 
+@pytest.mark.parametrize("force_path_style", [False, True])
+def test_copy_object(sample_directory: BucketPrefixFixture, force_path_style: bool):
```
muddyfish, line 408: No test for using the object info returned. One should be added, or the return type should be changed

s3torchconnectorclient/python/src/s3torchconnectorclient/_mountpoint_s3_client.pyi
matthieu-d4r: Indeed, will return `None` instead, oversight

s3torchconnectorclient/python/tst/integration/test_mountpoint_s3_integration.py
matthieu-d4r: Updating with a `@fixture`

s3torchconnectorclient/python/tst/integration/test_mountpoint_s3_integration.py
matthieu-d4r: It wouldn't 🙂 tried and only `b''.join(obj)` does the job; updating

s3torchconnectorclient/python/tst/integration/test_mountpoint_s3_integration.py
```diff
@@ -404,6 +404,50 @@ def test_delete_object_invalid_bucket(
         )
 
 
+@pytest.fixture
+@pytest.mark.parametrize("force_path_style", [False, True])
```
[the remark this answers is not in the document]
matthieu-d4r: Interesting, good catch; will remove

s3torchconnectorclient/python/tst/integration/test_mountpoint_s3_integration.py
matthieu-d4r: Will return `None` instead

Issue #53, opened 2023-11-20
Support multi-process/multi-node sharding for `S3IterableDataset`


jamesbornholt:
We currently don't have a built in way to do sharding for `S3IterableDataset`, so every worker process in a `DataLoader` will see the same stream of objects. We should have a way to do this.

In the meantime, something like this from `torchdata` will work as a workaround:
```python
from s3torchconnector import S3IterableDataset
from torch.utils.data import DataLoader
from torch.utils.data.datapipes.iter import IterableWrapper

dataset = S3IterableDataset.from_prefix("s3://doc-example-bucket/", region="us-west-2")
dataset = IterableWrapper(dataset, deepcopy=False)
dataset = dataset.sharding_filter() # Use torchdata's sharding for iterable datasets
loader = DataLoader(dataset, num_workers=2)
```

cfregly:
Related pull request for Megatron:  https://github.com/NVIDIA/Megatron-LM/pull/729

jamesbornholt:
The torchdata `IterableWrapper` is being deprecated in a future release, but it will still be present in PyTorch core. I've updated the code example above to point to that instead.

IsaevIlya:
Changes was released as part of [v1.3.0](https://github.com/awslabs/s3-connector-for-pytorch/releases/tag/v1.3.0 )

noepionentrust:
👋  folks, you may want to edit the documentation at  https://github.com/awslabs/s3-connector-for-pytorch/blob/main/examples/Getting%20started%20with%20the%20Amazon%20S3%20Connector%20for%20PyTorch.ipynb

```Important: When combining multi-process data loading with S3IterableDataset, by default each worker process will get its own replica of the dataset, and so each training sample will be duplicated num_workers times by the DataLoader. This is very likely not the behavior you want. We're currently [tracking this issue on GitHub](https://github.com/awslabs/s3-connector-for-pytorch/issues/53). In the meantime, as a workaround, you can use torchdata to wrap an S3IterableDataset and then shard it across workers:```

matthieu-d4r:
Hi @noepionentrust, good catch: we will update the documentation in an upcoming revision. Thanks!

PR #243, opened 2024-10-18
Add support for distributed training to S3IterableDataset

IsaevIlya:
## Description
This change adds support for distributed training to the S3IterableDataset. With this enhancement, the dataset becomes aware of multiple-node and multi-process configurations. By passing `enable_sharding=True`, each worker will have access to its own subset of the entire dataset, enabling efficient parallel data loading and processing.

## Additional context
In a multi-worker setup, where multiple workers (processes or threads) are used for data loading and processing, the S3IterableDataset partitions the dataset among the workers when you explicitly passed `enable_sharding=True`. This partitioning allows for parallel and efficient data handling.

## Related items
[Support multi-process/multi-node sharding for S3IterableDataset](https://github.com/awslabs/s3-connector-for-pytorch/issues/53)

## Testing
Integration and unit test were added.

--------


s3torchconnector/tst/e2e/test_distributed_training.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+start_methods = _get_start_methods()
+
+import torch.distributed as dist
+
+
+def setup(unique_port, rank, world_size):
+    dist.init_process_group(
+        backend="gloo",
```
matthieu-d4r, line 30: Shall we test other backend too, such as NCCL?

IsaevIlya: There shouldn't be a difference in type of used backed, but more importantly NCCL backed requires GPU, which is not available on hosts used for running tests.

s3torchconnector/tst/e2e/test_distributed_training.py
```diff
@@ -0,0 +1,183 @@
...
+
+    # Collect the results from the queue
+    results = [result_queue.get() for _ in range(num_processes)]
+
+    # Combine all uris_seen from the results
+    combined_uris_seen = Counter()
+    for uris_seen in results:
+        combined_uris_seen.update(uris_seen)
+
+    # Check if each item in image_directory was seen exactly once
```
muddyfish, line 140: Comment misleading - checks all keys were seen, but not how many times

s3torchconnector/tst/unit/test_s3iterable_dataset.py
```diff
@@ -218,6 +218,298 @@ def test_from_prefix_seek_no_head():
     head_object.assert_not_called()
 
 
+@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
muddyfish: If there's a reason these cases were chosen, can we add more(?) comments as it looks fairly opaque right now.

s3torchconnector/tst/unit/test_s3iterable_dataset.py
```diff
@@ -218,6 +218,298 @@ def test_from_prefix_seek_no_head():
...
+
+    assert isinstance(dataset, S3IterableDataset)
+    _verify_dataset(
+        dataset, expected_keys, lambda data: data._get_object_info is not None
+    )
+
+
+@pytest.mark.parametrize(
+    "keys, prefix, expected_keys, worker_id, num_workers, rank, world_size",
+    [
```
muddyfish, line 329: Same here - it's not obvious how these tests were chosen

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## v1.x.x (TBD)
+* Add support of distributed training to S3IterableDataset
+
+### Breaking changes
```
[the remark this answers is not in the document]
IsaevIlya: It is a breaking change for our previous approach, which was not entirely correct. Because of that, I recommend to make new approach as default going forward.

s3torchconnector/src/s3torchconnector/s3iterable_dataset.py
```diff
@@ -133,6 +148,26 @@ def _get_transformed_object(self, bucket_key: S3BucketKeyData) -> Any:
...
+        self._shard_index = num_workers * self._rank + worker_id
+        self._shard_count = num_workers * self._world_size
+
+        if self._shard_index == 0 and self._shard_count == 1:
+            return map(
+                self._get_transformed_object,
+                self._get_dataset_objects(self._get_client()),
+            )
+
+        sharded_objects = (
```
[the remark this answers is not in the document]
IsaevIlya: I think pickling happens before DataLoader calls `__iter__` for the first time, otherwise all the workers will access the same `worker_info` from parent worker.

s3torchconnector/tst/e2e/conftest.py
```diff
@@ -18,41 +18,62 @@ def getenv(var: str, optional: bool = False) -> str:
...
         full_key = f"{self.prefix}{key}"
         self.s3.put_object(Bucket=self.bucket, Key=full_key, Body=contents, **kwargs)
         self.contents[full_key] = contents
 
-    def __getitem__(self, index):
-        return self.contents[index]
-
-    def __iter__(self):
-        return iter(self.contents)
+    def get_context_only(self):
```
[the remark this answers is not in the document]
IsaevIlya: It is needed for running multi node integration tests. We can't transfer s3 client between processes and we don't need it, we only need the list of objects created for running that integration test.

s3torchconnector/tst/e2e/conftest.py
```diff
@@ -71,13 +92,29 @@ def get_test_bucket_prefix(name: str) -> BucketPrefixFixture:
...
+    return _create_image_directory_fixture(NUM_IMAGES, IMAGE_SIZE, request.node.name)
+
+
+@pytest.fixture
+def image_directory_for_dp(request) -> BucketPrefixFixture:
+    # When conducting distributed training tests, be cautious about the number of files (images) in the test dataset.
+    # If the total number of images cannot be evenly divided by the number of workers,
+    # the DistributedSampler will duplicate a subset of the images across workers to ensure an equal
+    # distribution of data among all processes. This duplication of images can potentially invalidate or
+    # compromise the results of the distributed training test.
```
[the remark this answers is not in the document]
IsaevIlya: It would invalidate the test results for S3MapDataset. Tests will start to fail, because some files will be passed to different workers by DistributedSampler more then once. I don't want to add complicated logic to try to adjust to DistributedSampler behavior.

s3torchconnector/tst/e2e/test_distributed_training.py
```diff
@@ -0,0 +1,183 @@
...
+    return dataloader
+
+
+# Allow us to construct our datasets in tests with either from_prefix or from_objects.
+dataset_builders = (from_prefix, from_objects)
+
+# Allow us to construct dataloaders in test with either S3MapDataset or S3IterableDataset
+dataloader_builders = (dataloader_for_iterable, dataloader_for_map)
+
+num_workers_to_test = [1,]
```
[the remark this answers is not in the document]
IsaevIlya: Good catch, it should be [1, 2, 3]. I indecently committed local changes for debugging integration tests failure.

s3torchconnector/tst/e2e/test_distributed_training.py
```diff
@@ -0,0 +1,183 @@
...
+    start_method: str,
+    dataset_builder: Callable,
+    dataloader_builder: Callable,
+    num_workers: int,
+    num_processes: int,
+    image_directory_for_dp: BucketPrefixFixture,
+):
+    # Generate unique port number in range [2000:61000] based on the test name
+    # to ensure that different test workers would use different ports
+    unique_port = hash(request.node) % 60000 + 2000
```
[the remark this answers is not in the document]
IsaevIlya: Thanks, I didn't know that. Will have a look how to get a proper hash value here.

s3torchconnector/tst/unit/test_s3iterable_dataset.py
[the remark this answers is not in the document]
IsaevIlya: I will try to think about visual representation, if it could be done in one or two lines

s3torchconnector/tst/e2e/conftest.py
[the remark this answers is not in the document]
muddyfish: I'd rename this method to make it more obvious, though from the sound of it, it doesn't seem too easy to explain consisely

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
...
+* When using the S3IterableDataset with multiple workers, each worker will have access to a distinct subset
```
fuatbasik, line 5: isn't it the otherway around for regular datasets. It is not sharded by default and you shard explicitly.

fuatbasik: `When using a [IterableDataset](https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset) with [multi-process data loading](https://pytorch.org/docs/stable/data.html#multi-process-data-loading). The same dataset object is replicated on each worker process, and thus the replicas must be configured differently to avoid duplicated data. See [IterableDataset](https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset) documentations for how to achieve this.`


https://pytorch.org/docs/stable/data.html#iterable-style-datasets

fuatbasik: Also, by not touching the constructor, you can avoid changing so many different places in the code.

README.md
```diff
@@ -115,6 +115,33 @@ For example, assuming the following directory bucket name `my-test-bucket--usw2-
...
+Both S3IterableDataset and S3MapDataset can be used for this purpose.
+
+### S3IterableDataset
+
+The S3IterableDataset can be directly passed to PyTorch's DataLoader for parallel and distributed training. 
+In this scenario, different workers will have access to unique portions of the dataset, enabling data parallelism. 
+To ensure backward compatibility with previous behavior, the share_dataset_within_process flag has been introduced. 
+When set to True, workers within the same process will share the same list of training objects.
+```
+dataset = S3IterableDataset.from_prefix(DATASET_URI, region=REGION, share_dataset_within_process=False)
```
fuatbasik, line 131: If we are to follow what i recommend (have an additional method call for dataset to shard data across workers)
`dataset = dataset.sharding_filter()`

s3torchconnector/src/s3torchconnector/s3iterable_dataset.py
```diff
@@ -133,6 +148,42 @@ def _get_transformed_object(self, bucket_key: S3BucketKeyData) -> Any:
...
-            self._get_transformed_object, self._get_dataset_objects(self._get_client())
+        worker_id = 0
+        num_workers = 1
+        if not self._share_dataset_within_process:
+            worker_info = torch.utils.data.get_worker_info()
+            if worker_info is not None:
+                worker_id = worker_info.id
+                num_workers = worker_info.num_workers
+
+        # In a multi-process setting (e.g., distributed training), the dataset needs to be
```
fuatbasik, line 159: is this right comment syntax?

s3torchconnector/src/s3torchconnector/s3iterable_dataset.py
```diff
@@ -133,6 +148,42 @@ def _get_transformed_object(self, bucket_key: S3BucketKeyData) -> Any:
         )
 
     def __iter__(self) -> Iterator[Any]:
-        return map(
-            self._get_transformed_object, self._get_dataset_objects(self._get_client())
+        worker_id = 0
+        num_workers = 1
+        if not self._share_dataset_within_process:
```
fuatbasik, line 153: nit: why not handing the if `self._share_dataset_within_process` first as the base condition and have the else implemented below? reasoning about NOT conditions are harder.

s3torchconnector/src/s3torchconnector/s3iterable_dataset.py
```diff
@@ -133,6 +148,42 @@ def _get_transformed_object(self, bucket_key: S3BucketKeyData) -> Any:
...
+        # num_workers: The total number of worker threads/processes within the process.
+        #
+        # The _shard_index and _shard_count variables are computed based on the above values,
+        # and they determine which subset of the dataset objects should be processed by the
+        # current worker thread/process in the current process rank.
+        self._shard_index = num_workers * self._rank + worker_id
+        self._shard_count = num_workers * self._world_size
+
+        if self._shard_index == 0 and self._shard_count == 1:
+            return map(
```
fuatbasik, line 179: this is the base case right. what we have today. I think this should either be on top of this method or at the bottom. Whatever we are changing should go above or beyond it, but not around. It is really hard to reason what needs to hold for behaviour to stay the same.

s3torchconnector/src/s3torchconnector/s3iterable_dataset.py
```diff
@@ -133,6 +148,42 @@ def _get_transformed_object(self, bucket_key: S3BucketKeyData) -> Any:
...
+                self._get_transformed_object,
+                self._get_dataset_objects(self._get_client()),
+            )
+
+        sharded_objects = (
+            obj
+            for idx, obj in enumerate(self._get_dataset_objects(self._get_client()))
```
fuatbasik, line 186: wouldn't this materialise entire iterable dataset at the first iter call?

CHANGELOG.md
IsaevIlya: I was focused on `thus the replicas must be configured differently to avoid duplicated data` part, when was looking how others were implementing their version of IterableDataset. For example, Google implementation support multiple workers by default [dataflux_iterable_dataset](https://github.com/GoogleCloudPlatform/dataflux-pytorch/blob/main/dataflux_pytorch/dataflux_iterable_dataset.py#L169-L175). The same is true for [text_to_text](https://github.com/NVIDIA/NeMo/blob/05f75862169b1dc7f2641c54ebe5ab3f6f8451cc/nemo/collections/asr/data/text_to_text.py#L458C41-L469) from NVIDIA Nemo framework. Both of these implementations refer to PyTorch documentation on how you should develop [IterableDataset](https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset):
```
When a subclass is used with DataLoader, each item in the dataset will be yielded 
from the DataLoader iterator. When num_workers > 0, each worker process will 
have a different copy of the dataset object, so it is often desired to configure 
each copy independently to avoid having duplicate data returned from the workers.
```
It is not saying that support of sharding is mandatory, but rather that it is desired. 
Do you think it would be better to move `desired` behavior behind the lever? My intention was to make that behavior enabled by default, because one of the often reason to not use S3IterableDataset was that we duplicates data in multi worker/node environment.

fuatbasik (changes requested):
Hi @IsaevIlya ,

Thanks a lot for your changes. I think we are making two important decisions in this PR and I want to revise them. 
1/ Interface -- I am not sure changing how we are creating the dataset to change how it will behave is the right interface. AFAIK, Pytorch datasets are generally modified using method calls that are returning dataset back like:
```
dataset = S3IterableDataset.from_prefix(prefix_uri, region=region)
dataset = torchdata.datapipes.iter.IterableWrapper(dataset)
dataset = dataset.sharding_filter()
dataset = dataset.load_from_tar()
```
I recommend we follow precedence here and do implement something similar. 

2/The default behaviour. Well i am quite rusty in this area but IIRC, current datasets by default do share data between workers. With this change we are introducing the opposite as the default. Again, lets make sure we are following precedence here and not confusing the customers. I added some relevant links.

README.md
```diff
@@ -114,7 +115,35 @@ For example, assuming the following directory bucket name `my-test-bucket--usw2-
...
+```
+dataset = S3IterableDataset.from_prefix(DATASET_URI, region=REGION, enable_sharding=True)
+dataloader = DataLoader(dataset, num_workers=4)
+```
+When `enable_sharding` is set to True, the dataset will be automatically sharded across available number of workers. 
+This sharding mechanism supports both parallel training on a single host and distributed training across multiple hosts. 
+Each worker, regardless of its host, will load and process a distinct subset of the dataset.
+### S3MapDataset
+
+For the S3MapDataset, you need to pass it to DataLoader along with a DistributedSampler wrapped around it.
```
fuatbasik, line 139: nit: hyperlink to DistributedSampler?

s3torchconnector/src/s3torchconnector/s3iterable_dataset.py
```diff
@@ -133,6 +148,47 @@ def _get_transformed_object(self, bucket_key: S3BucketKeyData) -> Any:
...
+        if self._shard_count > 1:
+            # we have more than one shard, so need to distribute dataset between shards
+            sharded_objects = (
+                obj
+                for idx, obj in enumerate(self._get_dataset_objects(self._get_client()))
+                if idx % self._shard_count == self._shard_index
+            )
+            return map(self._get_transformed_object, sharded_objects)
+
+        # only one shard, so return the entire dataset
```
fuatbasik, line 190: nit: or sharding is not enabled

s3torchconnector/tst/e2e/conftest.py
```diff
@@ -18,41 +18,68 @@ def getenv(var: str, optional: bool = False) -> str:
...
         """Upload an S3 object to this prefix of the bucket."""
         full_key = f"{self.prefix}{key}"
         self.s3.put_object(Bucket=self.bucket, Key=full_key, Body=contents, **kwargs)
         self.contents[full_key] = contents
 
-    def __getitem__(self, index):
-        return self.contents[index]
+    def get_data_snapshot(self):
+        """
+        Returns a read-only copy of the current instance's data.
```
fuatbasik, line 75: nit: should comments start at the same line with """ or at the next line? Methods above/below are inconsistent.

s3torchconnector/tst/e2e/conftest.py
```diff
@@ -71,13 +98,29 @@ def get_test_bucket_prefix(name: str) -> BucketPrefixFixture:
...
-    fixture = get_test_bucket_prefix(f"{request.node.name}/image_directory")
-    for i in range(NUM_IMAGES):
-        data = np.random.randint(0, 256, IMAGE_SIZE * IMAGE_SIZE * 3, np.uint8)
-        data = data.reshape(IMAGE_SIZE, IMAGE_SIZE, 3)
+    return _create_image_directory_fixture(NUM_IMAGES, IMAGE_SIZE, request.node.name)
+
+
+@pytest.fixture
+def image_directory_for_dp(request) -> BucketPrefixFixture:
+    # When conducting distributed training tests, be cautious about the number of files (images) in the test dataset.
```
fuatbasik, line 108: nit: not sure if we used this types of comments `with #` anywhere else.

fuatbasik (approved):
Thanks a lot @IsaevIlya . Thank you for addressing my comments. I put some nitpicks but nothing blocking.

s3torchconnector/tst/e2e/test_common.py
```diff
@@ -0,0 +1,36 @@
...
+#  // SPDX-License-Identifier: BSD
+
+import platform
+import torch
+from s3torchconnector import S3Reader
+
+from typing import Tuple
+
+
+def _get_fork_methods() -> set:
```
muddyfish, line 11: Nit: `set[str]` or `Set[str]`

s3torchconnector/tst/unit/test_s3iterable_dataset.py
```diff
@@ -218,6 +218,387 @@ def test_from_prefix_seek_no_head():
     head_object.assert_not_called()
 
 
+"""
```
muddyfish, line 221: Would some property based tests be useful here? 

Expected_keys is effectively calculated as
```
all_keys[(worker_id*num_workers)+rank::world_size*num_workers]
```

s3torchconnector/src/s3torchconnector/s3iterable_dataset.py
```diff
@@ -133,6 +148,47 @@ def _get_transformed_object(self, bucket_key: S3BucketKeyData) -> Any:
...
+
+        worker_id: The ID of the current worker thread/process within the process.
+        num_workers: The total number of worker threads/processes within the process.
+
+        The _shard_index and _shard_count variables are computed based on the above values,
+        and they determine which subset of the dataset objects should be processed by the
+        current worker thread/process in the current process rank.
+        """
+
+        self._shard_index = num_workers * self._rank + worker_id
```
rajdchak, line 178: are we assuming that number of workers per rank would be equal everytime? If the number of workers are not equal per rank, I think it can lead to issues while distributing the data.

IsaevIlya: good question! yes, we are assuming uniform distribution, but let me think, if we can make it more agile and support  uneven distribution

s3torchconnector/tst/unit/test_s3iterable_dataset.py
IsaevIlya: Replaced with property based, but have to change what is tested. Now it test that all keys are evenly distributed between ranks without duplication.

s3torchconnector/tst/unit/test_s3iterable_dataset.py
```diff
@@ -218,6 +220,140 @@ def test_from_prefix_seek_no_head():
...
+        create_from_prefix (bool): Whether to create the dataset from a prefix or a list of object URIs.
+        world_size_and_num_workers (tuple): A tuple containing world_size and num_workers for each rank.
+    """
+    prefix1 = "obj"
+    prefix2 = "test"
+    world_size, num_workers_per_rank = world_size_and_num_workers
+
+    # Assume valid input combinations
+    assume(keys_for_prefix1 and keys_for_prefix2)
+    assume(world_size >= 1 and world_size == len(num_workers_per_rank))
```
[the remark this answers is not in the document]
IsaevIlya: It should be assume. We want to skip invalid input parameters, rather then fail on them.

s3torchconnector/tst/unit/test_s3iterable_dataset.py
```diff
@@ -218,6 +220,140 @@ def test_from_prefix_seek_no_head():
...
+        assert len(set(flattened_keys_from_workers)) == len(
+            all_keys
+        ), "All keys should be processed"
+        assert set(flattened_keys_from_workers) == set(
+            all_keys
+        ), "Union of keys should be the same"
+        expected_num_per_rank = len(all_keys) // world_size
+    assert all(
+        [abs(num_keys - expected_num_per_rank) <= 1 for num_keys in num_keys_per_rank]
+    ), "The number of keys should be evenly distributed across ranks, with a difference of at most 1"
```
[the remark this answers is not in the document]
IsaevIlya: It is to ensure that we don't have situation, when the entire dataset was actually processed by single rank.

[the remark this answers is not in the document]
muddyfish: Then why do you need the `abs` call?

s3torchconnector/tst/unit/test_s3iterable_dataset.py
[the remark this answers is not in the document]
muddyfish: From what I understood, the method you're generating these parameters should always have these conditions hold. Definitely up to you if you want to keep it this way, but I was thinking it might be worth changing to assert so you know the property always holds

s3torchconnector/tst/unit/test_s3iterable_dataset.py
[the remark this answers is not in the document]
IsaevIlya: Yes, you are right. We don't need `abs` here.

matthieu-d4r:
LGTM!

matthieu-d4r (approved):

README.md
```diff
@@ -178,8 +208,11 @@ If you discover a potential security issue in this project we ask that you notif
...
 This project has adopted the [Amazon Open Source Code of Conduct](https://aws.github.io/code-of-conduct).
-See [CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md) for more details.
+See [CODE_OF_CONDUCT.md](https://github.com/awslabs/s3-connector-for-pytorch/blob/main/doc/CODE_OF_CONDUCT.md) for 
+more details.
 
 ## License
 
-Amazon S3 Connector for PyTorch has a BSD 3-Clause License, as found in the [LICENSE](LICENSE) file.
+Amazon S3 Connector for PyTorch has a BSD 3-Clause License, as found in the 
+[LICENSE](https://github.com/awslabs/s3-connector-for-pytorch/blob/main/LICENSE) file.
```
matthieu-d4r, line 217: Unimportant question: why replacing relative with absolute?

matthieu-d4r (approved):

PR #246, opened 2024-10-23
ci: tweak the Python workflows

matthieu-d4r:
Remove manual installations of PyTorch everywhere, as the latter should
be handled while installing the library through `pip`. Remove `-e` flag
while installing our packages, and refer to them with `./`. Remove
`s3torchbenchmarking` unit tests runs, unnecessary for now. Reformat
some workflow files (cosmetic).

Drop support (in worfklows) for macOS x86_64 architecture, since PyTorch
dropped support to since v2.3
(https://dev-discuss.pytorch.org/t/pytorch-macos-x86-builds-deprecation-starting-january-2024/1690),
and future DCP implementations will require that version.

--------

### DIFF
--- a/.github/workflows/ci-integration-pr.yml
+++ b/.github/workflows/ci-integration-pr.yml
@@ -2,7 +2,7 @@ name: Integration tests (PR)
 
 on:
   pull_request_target:
-    branches: [ "main", "feature/*", "ci/*" ]
+    branches: [ "main", "feature/*" ]
 
 permissions:
   id-token: write
--- a/.github/workflows/ci-integration-pr.yml
+++ b/.github/workflows/ci-integration-pr.yml
@@ -2,7 +2,7 @@ name: Integration tests (PR)
 
 on:
   pull_request_target:
-    branches: [ "main", "feature/*" ]
+    branches: [ "main", "feature/*", "ci/*" ]
 
 permissions:
   id-token: write
--- a/.github/workflows/python-checks.yml
+++ b/.github/workflows/python-checks.yml
@@ -42,8 +42,9 @@ jobs:
           cache: 'pip'
       - name: Install Python dependencies
         run: |
-          pip install "./s3torchconnectorclient[test]"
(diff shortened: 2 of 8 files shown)


.github/workflows/python-checks.yml
```diff
@@ -43,23 +43,18 @@ jobs:
...
+          python -m pip install "./s3torchconnectorclient[test]"
+          python -m pip install "./s3torchconnector[test]"
 
-      - name: s3torchconnectorclient unit tests
+      - name: Run s3torchconnectorclient unit tests
         run: pytest s3torchconnectorclient/python/tst/unit --hypothesis-profile ci --hypothesis-show-statistics -c ./
-      - name: s3torchconnector unit tests
+      - name: Run s3torchconnector unit tests
         run: pytest s3torchconnector/tst/unit --ignore s3torchconnector/tst/unit/lightning --hypothesis-profile ci --hypothesis-show-statistics -c ./
-      - name: s3torchbenchmarking unit tests
```
IsaevIlya, line 56: Do we have any conflicts because of s3torchbenchmarking unit tests? I know that they essentially do nothing at the moment, but if we will delete them now, we can forget to add them back later.

.github/workflows/ci-integration-pr.yml
```diff
@@ -2,7 +2,7 @@ name: Integration tests (PR)
 
 on:
   pull_request_target:
-    branches: [ "main", "feature/*" ]
```
IsaevIlya, line 5: Why we need `ci` prefix here?

s3torchconnectorclient/pyproject.toml
```diff
@@ -32,7 +32,8 @@ test = [
     "flake8",
     "black",
     "mypy",
-    "Pillow"
+    "numpy",
```
IsaevIlya, line 35: Why do we need `numpy`?

matthieu-d4r:
Cancelling with PR: will include changes within the `feature/dcp` branch.

PR #248, opened 2024-10-27
feat(dcp): add implementations for FSDP support

matthieu-d4r:
Add `S3FileSystem`, `S3StorageWriter`, and `S3StorageReader` classes. Add relating unit and e2e tests.

--------


s3torchconnector/src/s3torchconnector/dcp/fsdp_filesystem.py
```diff
@@ -0,0 +1,249 @@
...
+        try:
+            parse_s3_uri(_path_to_str(checkpoint_id))
+        except ValueError:
+            return False
+        return True
+
+    @retry(
+        retry=retry_if_exception_type(S3Exception),
+        stop=stop_after_attempt(3),
+        wait=wait_fixed(3) + wait_random(0, 2),
```
rajdchak, line 179: should we go with exponential backoff strategy?

s3torchconnector/src/s3torchconnector/dcp/fsdp_filesystem.py
```diff
@@ -0,0 +1,249 @@
...
+        pass
+
+    def exists(self, path: Union[str, os.PathLike]) -> bool:
+        logger.debug("exists %s", path)
+
+        path = _path_to_str(path)
+        bucket, key = parse_s3_uri(path)
+        try:
+            self._client.head_object(bucket, key)
+        except S3Exception:
```
rajdchak, line 149: should we be returning a false in all cases of S3 Exception, or return a false only if the error thrown by Mountpoint S3Client is ```NotFound``` and in other cases just raise that exception?

s3torchconnector/src/s3torchconnector/dcp/fsdp_filesystem.py
```diff
@@ -0,0 +1,249 @@
...
+            path (Union[str, os.PathLike]): The S3 path to read or write.
+            mode (str): The mode for the stream. Supports 'rb' for read mode and 'wb' for write mode.
+
+        Yields:
+            io.BufferedIOBase: A stream for reading or writing to S3.
+
+        Raises:
+            ValueError: If the mode is not 'rb' or 'wb'.
+        """
+        path = _path_to_str(path)
```
rajdchak, line 55: Nit: Here and in other places, may be we can avoid reassigning to the same variable.

s3torchconnector/src/s3torchconnector/dcp/fsdp_filesystem.py
```diff
@@ -0,0 +1,249 @@
...
+            raise ValueError(
+                f"Source and destination buckets cannot be different (rename does not support cross-buckets operations)"
+            )
+
+        self._client.copy_object(
+            src_bucket=old_bucket,
+            src_key=old_key,
+            dst_bucket=new_bucket,
+            dst_key=new_key,
+        )
```
rajdchak, line 135: should we have some success log after the completion of each operation like copy and then delete? It can be easy for debugging.

s3torchconnector/src/s3torchconnector/dcp/fsdp_filesystem.py
matthieu-d4r: Good point: we can restrict to `NotFound`, and let everything surface to the client; will change

s3torchconnector/src/s3torchconnector/dcp/fsdp_filesystem.py
matthieu-d4r: Why not: would you recommend any parameters? Like, I can think of:
```
@retry(
    retry=retry_if_exception_type(S3Exception),
    stop=stop_after_attempt(3),
    wait=wait_random_exponential(multiplier=1, max=5),
    before_sleep=before_sleep_log(logger, logging.WARNING),
    after=after_log(logger, logging.ERROR),
    reraise=True,
)
```

rajdchak: Also, should we do the retry for some specific scenarios. In this case, it looks like we are retrying for all errors, irrespective of whether it is retryable or not.

matthieu-d4r: That's true: I can think of:
1. [`Request throttled`](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-client/src/s3_crt_client.rs#L1030-L1032)
2. [`Failed to construct request`](https://github.com/awslabs/mountpoint-s3/blob/main/mountpoint-s3-client/src/s3_crt_client.rs#L1006-L1008)

WDYT?

matthieu-d4r:
Further comments to be made under https://github.com/awslabs/s3-connector-for-pytorch/pull/251 (this PR has an incorrect destination branch).

PR #259, opened 2024-11-12
feat(dcp): add benchmark capabilities for DCP

matthieu-d4r:
Create a dedicated `dcp` Python package within `s3torchbenchmarking`, to run benchmarks against fsspec.

--------


s3torchbenchmarking/src/s3torchbenchmarking/dcp/benchmark.py
```diff
@@ -0,0 +1,154 @@
...
+    )
+
+
+def validate_world_size(world_size: int) -> int:
+    """Enforce `world_size` to be within the current node's capacity.
+
+    FIXME: only works when the backend is "nccl"; what about "gloo" and CPUs?
+    """
+    device_count = torch.cuda.device_count()
+    if world_size > device_count:
```
rajdchak, line 87: I don't think we should restrict this, world size can be greater than number of gpus, there is no limitation like this.

s3torchbenchmarking/src/s3torchbenchmarking/dcp/benchmark.py
```diff
@@ -0,0 +1,154 @@
...
+    storage_writer: FileSystemWriter,
+    save_timestamps: Queue,
+) -> None:
+    """Execute the actual code for checkpoint saving.
+
+    This function is meant to be executed in subprocesses."""
+
+    # FIXME: configure logging in subprocess accordingly
+
+    torch.cuda.set_device(rank)
```
rajdchak, line 139: For world size greater than gpus this throws error.  Instead we should set this as follows
```
rank = dist.get_rank()
    device_id = rank % torch.cuda.device_count()
    torch.cuda.set_device(device_id)
```

s3torchbenchmarking/src/s3torchbenchmarking/dcp/benchmark.py
```diff
@@ -0,0 +1,156 @@
...
+    setup(backend=backend, world_size=world_size, rank=rank)
+    rank = dist.get_rank()
+
+    device_id = rank % torch.cuda.device_count()
+    model.to(device_id)
+
+    ddp_model = DistributedDataParallel(model, device_ids=[device_id])
+
+    start_time = perf_counter()
+    for i in range(epochs):
```
rajdchak, line 149: Adding a loop here does not make much sense to me, we should instead add the epochs in the place where the we are spawning as we are trying to find a mean run time over 5 epochs.

s3torchbenchmarking/src/s3torchbenchmarking/dcp/benchmark.py
```diff
@@ -0,0 +1,156 @@
...
+        dcp.save(ddp_model.state_dict(), storage_writer=storage_writer)
+    end_time = perf_counter()
```
rajdchak, line 151: Let's think about edge cases, the start and end time here is still impacted by the load time of the model and other processes going on before this, we should remove the time taken for all the processes happening before this step, one approach I tried is

def run(
    rank: int,
    model: Module,
    world_size: int,
    storage_writer: FileSystemWriter,
    shared_start_times: list,
    shared_end_times: list,
):
    start_time = perf_counter()
    logger.info("Running on rank %s...", rank)

    #torch.cuda.set_device(rank)
    setup(rank, world_size)
    rank = dist.get_rank()
    device_id = rank % torch.cuda.device_count()
    torch.cuda.set_device(device_id)

    logger.debug("Device ID: %s", device_id)

    model.to(device=torch.device("cpu"))
    wrapped_model = DistributedDataParallel(model)

    logger.info("Saving checkpoint on rank %s...", rank)

    model_load_end = perf_counter()

    dcp.save(
        wrapped_model.state_dict(),
        storage_writer=storage_writer,
    )
    end_time = perf_counter() - (model_load_end - start_time)

s3torchbenchmarking/src/s3torchbenchmarking/dcp/benchmark.py
```diff
@@ -0,0 +1,156 @@
...
+
+    manager = multiprocessing.Manager()
+    save_timestamps: Queue[float] = manager.Queue()
+
+    processing_times = Distribution()
+    ptp_save_times = Distribution()  # "ptp" == "peak to peak", or "diff(max, min)"
+
+    with ResourceMonitor() as monitor:
+        start_time = perf_counter()
+        multiprocessing.spawn(
```
rajdchak, line 51: let us run this for 5 epochs, this will help to find a mean value of the run time

s3torchbenchmarking/src/s3torchbenchmarking/dcp/benchmark.py
matthieu-d4r: Got it, lifting that up

s3torchbenchmarking/src/s3torchbenchmarking/dcp/benchmark.py
matthieu-d4r: I tried that and got the following error:

```
torch.distributed.DistBackendError: NCCL error in: /opt/conda/conda-bld/pytorch_1716905969073/work/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp:1970, invalid usage (run with NCCL_DEBUG=WARN for details), NCCL version 2.20.5
ncclInvalidUsage: This usually reflects invalid usage of NCCL library.
Last error:
Duplicate GPU detected : rank 4 and rank 0 both on CUDA device 1b0
```

(with `world_size` of 8 on g5.12xlarge)
😢

s3torchbenchmarking/src/s3torchbenchmarking/dcp/README.md
```diff
@@ -0,0 +1,134 @@
+## PyTorch's Distributed Checkpoint (DCP) benchmarks
+
+The `dcp` Python package holds all the logic to execute benchmarks for [PyTorch's Distributed Checkpointing][DCP]
+feature against the `s3torchconnector` library.
+
+### Purpose
+
+The benchmarks are designed to test the "save" and "load" mechanisms of PyTorch
```
IsaevIlya, line 8: nit: we could omit "load" part, as we are only providing way to measure "save" part

s3torchbenchmarking/src/s3torchbenchmarking/dcp/README.md
```diff
@@ -0,0 +1,134 @@
...
+(`torch.distributed.checkpoint[save,load]`), and compare the s3torchconnector library performance vs. other libraries,
+like fsspec.
```
IsaevIlya, line 10: lets not mention fsspec directly, but better leave "other libraries and local storage"

s3torchbenchmarking/src/s3torchbenchmarking/dcp/benchmark.py
```diff
@@ -0,0 +1,126 @@
...
+    model: Module,
+    storage_writer: FileSystemWriter,
+    save_durations: Queue,
+) -> None:
+    """Execute the actual code for checkpoint saving.
+
+    This function is meant to be executed in subprocesses."""
+
+    setup(cfg.backend, world_size=cfg.world_size, rank=rank)
+    rank = dist.get_rank()
```
IsaevIlya, line 108: why do we need to call `dist.get_rank()` here, if we already have `rank`?

s3torchbenchmarking/src/s3torchbenchmarking/dcp/config.yaml
```diff
@@ -0,0 +1,16 @@
+s3:
+  region: ???
+  uri: ???
+epochs: 4
+path: nvme # only used when `checkpoint.storage` contains `disk`, ignored for `s3`
```
IsaevIlya, line 5: Can we reuse configs from `conf/checkpoint/*` to define configuration for disk and s3?

IsaevIlya: Additionally, can we move this config to `s3torchbenchmarking/conf` directory, that all configs were located in one place.

s3torchbenchmarking/src/s3torchbenchmarking/dcp/models.py
```diff
@@ -0,0 +1,61 @@
...
+    ),
+    # ~1.7 GB model
+    "small-2": BenchmarkModel(
+        CLIPModel.from_pretrained, "openai/clip-vit-large-patch14"
+    ),
+    # ~12 GB model
+    "medium": BenchmarkModel(AutoModelForSeq2SeqLM.from_pretrained, "bigscience/T0_3B"),
+    # ~45 GB model
+    "large": BenchmarkModel(AutoModelForSeq2SeqLM.from_pretrained, "bigscience/T0pp"),
+}
```
IsaevIlya, line 54: Can we use the same name for models that we use for Lightning benchmark? For example, not "large", but "t0pp".

s3torchbenchmarking/src/s3torchbenchmarking/dcp/models.py
IsaevIlya, line 1: I'm seeing a duplicated logic with https://github.com/awslabs/s3-connector-for-pytorch/blob/main/s3torchbenchmarking/src/s3torchbenchmarking/lightning_benchmark.py
Could we reuse the common part in both places?

s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
```diff
@@ -196,33 +196,24 @@ class S3StorageWriter(FileSystemWriter):
     def __init__(
         self,
         region: str,
-        path: Union[str, os.PathLike],
-        single_file_per_rank: bool = True,
-        thread_count: int = 1,
-        per_thread_copy_ahead: int = 10_000_000,
-        overwrite: bool = False,
+        uri: str,
+        **kwargs,
```
IsaevIlya, line 200: Nice!

s3torchbenchmarking/src/s3torchbenchmarking/dcp/results.py
IsaevIlya, line 1: Could it be integrated with https://github.com/awslabs/s3-connector-for-pytorch/blob/main/s3torchbenchmarking/src/s3torchbenchmarking/hydra_callbacks.py ?

s3torchbenchmarking/src/s3torchbenchmarking/dcp/benchmark.py
```diff
@@ -0,0 +1,126 @@
...
+        torch.cuda.set_device(device_id)
+    else:
+        device_id = rank % torch.cpu.device_count()
+        torch.cpu.set_device(device_id)
+
+    model.to(device_id)
+    ddp_model = DistributedDataParallel(model, device_ids=[device_id])
+
+    start_time = perf_counter()
+    for i in range(cfg.epochs):
```
rajdchak, line 120: Adding a loop here does not make much sense to me, we should instead add the epochs in the place where the we are spawning as we are trying to find a mean run time over 5 epochs.

s3torchbenchmarking/src/s3torchbenchmarking/dcp/benchmark.py
matthieu-d4r: Something I saw in the benchmark examples on PyTorch, but indeed does not make sense (and runs fine w/o it): removing

s3torchbenchmarking/src/s3torchbenchmarking/dcp/config.yaml
matthieu-d4r: My overall approach for this CR was to bundle everything under `dcp`, so it lives on its own folder/package; will move things around

s3torchbenchmarking/src/s3torchbenchmarking/dcp/models.py
matthieu-d4r: Yes: I didn't want to have some dependencies on Lightning, as I believe those two should live separately and be independent. I also think a further refactoring of some of the Lightning utils could be done to move and share them with other benchmarks, but out of scope for this CR.

s3torchbenchmarking/src/s3torchbenchmarking/dcp/results.py
matthieu-d4r: I chose to opt-out from callbacks because it is currently an experimental features of Hydra...

s3torchbenchmarking/src/s3torchbenchmarking/dcp/benchmark.py
matthieu-d4r: That was @IsaevIlya suggestions... the setup you describe was what I had before; let's discuss that tomorrow (Wed. 12th) 🙂

s3torchbenchmarking/src/s3torchbenchmarking/dcp/benchmark.py
matthieu-d4r: See my comment other comment (below).

IsaevIlya (approved):

rajdchak (approved):

PR #261, opened 2024-11-18
Update documentation for supporting DCP

IsaevIlya:
## Description
Update documentation to include information about DCP support

--------


README.md
```diff
@@ -114,6 +114,64 @@ For example, assuming the following directory bucket name `my-test-bucket--usw2-
...
+PyTorch 2.3 or newer is required.
+To use the distributed checkpoints feature, install S3 Connector for PyTorch with the `dcp` extra:
+
+```sh
+pip install s3torchconnector[dcp]
+```
+
+### Sample Example
+
+End to end examples for using distributed checkpoints with S3 Connector for PyTorch
```
matthieu-d4r, line 141: nit: "End to end" in this context is usually hyphenated: "End-to-end" ([source](https://en.wikipedia.org/wiki/End-to-end))

examples/dcp/stateful_example.py
```diff
@@ -0,0 +1,138 @@
+# based on PyTorch sample https://github.com/pytorch/pytorch/blob/main/torch/distributed/checkpoint/examples/stateful_example.py
```
matthieu-d4r, line 1: should we add the Amazon copyright header?

examples/dcp/stateful_example.py
```diff
@@ -0,0 +1,138 @@
...
+            if not torch.allclose(param1, param2, rtol=rtol, atol=atol):
+                print(f"Parameters don't match for {name1}. Rank:{rank}")
+                print(f"Max difference: {(param1 - param2).abs().max().item()}. Rank:{rank}")
+                return False
+
+    print("All parameters match within the specified tolerance. Rank:{rank}")
+    return True
+
+def run(rank, world_size, device="cuda"):
+    # Set up world pg
```
matthieu-d4r, line 85: "pg"?

examples/dcp/stateful_example.py
```diff
@@ -0,0 +1,138 @@
...
+    os.environ["MASTER_ADDR"] = "localhost"
+    os.environ["MASTER_PORT"] = "12355"
+
+    dist.init_process_group("cpu:gloo,cuda:nccl", rank=rank, world_size=world_size)
```
matthieu-d4r, line 89: What does `"cpu:gloo,cuda:nccl"` mean?

examples/dcp/stateful_example.py
```diff
@@ -0,0 +1,138 @@
...
+    torch.cuda.set_device(rank)
+
+    print(f"Train initial model on rank:{rank}")
+    model, optim = _init_model(device, world_size)
+    _train(model, optim, train_steps=2)
+
+    region = "us-east-1"
+    s3_uri = f"s3://PLEASE-PROVIDE-ACTUAL-BUCKET-NAME/"
```
matthieu-d4r, line 97: nit: could be uppercase variables at the top of the script, so user would know from the top they need to fill that up; also, unnecessary `f` at the front

examples/dcp/stateful_example.py
```diff
@@ -0,0 +1,138 @@
...
+        storage_writer=storage_writer,
+    )
+
+    # presumably do something else
+    print(f"Train another model on rank:{rank}")
+    loaded_model, loaded_optim = _init_model(device, world_size)
+    _train(loaded_model, loaded_optim, train_steps=4)
+
+    print(f"Check that models are different on rank:{rank}")
+    assert _compare_models(model, loaded_model, rank) is False
```
matthieu-d4r, line 113: nit: `assert not ...`?

examples/dcp/stateful_example.py
```diff
@@ -0,0 +1,138 @@
...
+    def forward(self, x):
+        return self.net4(self.net3(self.net2(self.net1(x))))
+
+    def get_input(self):
+        return torch.rand(8, 8, device="cuda")
+
+
+def _make_stateful(model, optim):
+    _patch_model_state_dict(model)
+    _patch_optimizer_state_dict(model, optimizers=optim)
```
matthieu-d4r, line 38: I remember in PyTorch's examples a class called `AppState`, which I think does what you're doing here: what about using it instead of those two private functions, so the user could relate better to it?

In fact, why not use https://pytorch.org/tutorials/recipes/distributed_checkpoint_recipe.html#saving instead, for simplicity?

matthieu-d4r (approved):
No blockers, a few nits and fewer non-nits

examples/dcp/stateful_example.py
IsaevIlya: That is complicated question, as the source of that example belong to PyTorch. Let me try to find how we dealt with that in other cases.

examples/dcp/stateful_example.py
IsaevIlya: It should be process group

examples/dcp/stateful_example.py
IsaevIlya: This is telling PyTorch to use `gloo` backend for training part processed on CPU and to use `nccl` backend for training part processed on GPU.

examples/dcp/stateful_example.py
IsaevIlya: Agree with removing `f`. Regarding location, I think it is better to keep it close to first usage of `S3StorageWriter`, that reader won't be needed to jump around file for looking what these variables mean.

IsaevIlya: I changed it to be an method's parameters to be on the same line with lighting samples

examples/dcp/stateful_example.py
IsaevIlya: I think this example shows more low level approach that is easier to tune up. However, samples that you mentioned easier to understand, I will add them as separate samples.

rajdchak (commented):
Can we combine the dcp_save and dcp_load examples in one file? I see a lot of code is repeated in both the files.

rajdchak (commented):
Or, we can move all the common things in one utils file from which all the examples file can use.

IsaevIlya:
> Can we combine the dcp_save and dcp_load examples in one file? I see a lot of code is repeated in both the files.
> Or, we can move all the common things in one utils file from which all the examples file can use.

While reducing code duplication is generally good practice, I intentionally kept these examples separate. Main reason is standalone demonstrations. Each example is designed to be self-contained, allowing users to understand and run either the save or load functionality independently without needing to reference other files. While moving common code to one place is a valid approach for production code, which I totally support, for these educational examples, the benefits of keeping each example self-contained looks more important to me then code duplication.

examples/dcp/stateful_example.py
```diff
@@ -0,0 +1,147 @@
...
+        state_dict={"model": loaded_model, "optimizer": loaded_optim},
+        storage_reader=storage_reader,
+    )
+    print(f"Check that loaded model and original model are the same on rank:{rank}")
+    assert _compare_models(model, loaded_model, rank)
+
+    print(f"Train loaded model on rank:{rank}")
+    _train(loaded_model, loaded_optim, train_steps=2)
+
+    print(f"Quiting on rank:{rank}")
```
rajdchak, line 134: The run method is too big, can you break it down into smaller methods for the ease of readability and understanding?

rajdchak (approved):

PR #276, opened 2024-12-04
Update user agent for DCP

IsaevIlya:
## Description
Update user agent for DCP

--------

### DIFF
--- a/s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
+++ b/s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
@@ -34,8 +34,12 @@ logger = logging.getLogger(__name__)
 class S3FileSystem(FileSystemBase):
     def __init__(self, region: str, s3_client: Optional[S3Client] = None) -> None:
         self._path: Union[str, os.PathLike] = ""
-        self.user_agent = UserAgent(["dcp", torch.__version__])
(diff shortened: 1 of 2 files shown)


s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
```diff
@@ -22,17 +22,20 @@
...
+from .._user_agent import UserAgent
 
 logger = logging.getLogger(__name__)
 
 
 class S3FileSystem(FileSystemBase):
     def __init__(self, region: str, s3_client: Optional[S3Client] = None) -> None:
         self._path: Union[str, os.PathLike] = ""
-        self._client = s3_client if s3_client is not None else S3Client(region)
+        self.user_agent = UserAgent(["dcp", torch.__version__])
```
matthieu-d4r, line 37: Do we need to have a `self.user_agent` field? Or could we just pass it to the `S3Client` below? E.g.:

```python
self._client = s3_client if s3_client else S3Client(region=region, user_agent=UserAgent(["dcp", torch.__version__]))
```

Also, shall we unit test this?

IsaevIlya: Made `user_agent` as variable and added tests

matthieu-d4r (approved):
You added e2e tests, while I thought about unit tests 🙂 LGTM anyway, approving

PR #280, opened 2024-12-10
refactor(benchmarks): improve jobs and runs saves, add write to DynamoDB

matthieu-d4r:
Create a "constants.py" file. Improve Hydra's callback behaviour and
save run results to DynamoDB (optionally). Improve shell scripts used to
run the benchmarks. Add a script to prepare an EC2 instance for
benchmarking.

--------

### DIFF
--- a/s3torchbenchmarking/utils/prepare_ec2_instance.sh
+++ b/s3torchbenchmarking/utils/prepare_ec2_instance.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+#
+# Script to prepare an EC2 instance for PyTorch benchmarks.
+
+set -euo pipefail
+
+# Sanity check
+sudo yum -y upgrade # OR, `sudo apt -y upgrade`
+
+# Activate the default PyTorch env
+#
+# This command only works on specific AMI (DL-AMI): make sure to check the compatibility between the latter and the EC2
+# instance type currently in use (if incompatible, the command will either fail or generate a warning).
+source activate pytorch
+
+# Install Rust (https://www.rust-lang.org/learn/get-started)
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+. "$HOME/.cargo/env"
+
+# Addresses error "RuntimeError: operator torchvision::nms does not exist" while trying the run the benchmarks
+conda install -y pytorch torchvision torchaudio pytorch-cuda=12.4 -c pytorch -c nvidia
+
+# Addresses error "TypeError: canonicalize_version() got an unexpected keyword argument 'strip_trailing_zero'" while
+# trying to install s3torchbenchmarking
+pip install "setuptools<71"
+
+# Required to build s3torchconnectorclient (see its README)
+sudo yum -y install cmake3 clang
+
+pip install 's3torchconnector[lightning,dcp]'
+cd s3torchbenchmarking && pip install -e .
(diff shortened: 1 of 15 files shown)


s3torchbenchmarking/utils/prepare_ec2_instance.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Addresses error "TypeError: canonicalize_version() got an unexpected keyword argument 'strip_trailing_zero'" while
+# trying to install s3torchbenchmarking
+pip install "setuptools<71"
+
+# Required to build s3torchconnectorclient (see its README)
+sudo yum -y install cmake3 clang
+
+pip install 's3torchconnector[lightning,dcp]'
+cd s3torchbenchmarking && pip install -e .
```
IsaevIlya, line 31: Do we need to "git clone" project first?

s3torchbenchmarking/utils/prepare_ec2_instance.sh
```diff
@@ -0,0 +1,31 @@
...
+sudo yum -y upgrade # OR, `sudo apt -y upgrade`
+
+# Activate the default PyTorch env
+#
+# This command only works on specific AMI (DL-AMI): make sure to check the compatibility between the latter and the EC2
+# instance type currently in use (if incompatible, the command will either fail or generate a warning).
+source activate pytorch
+
+# Install Rust (https://www.rust-lang.org/learn/get-started)
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
IsaevIlya, line 17: why do we need rust if we install s3torchconnectorclient from pip?

IsaevIlya (approved):
I generally fine with the changes. Left few minor questions. Also, could you explain how we will able to connect results from different EC2 for the same benchmark run?

s3torchbenchmarking/utils/prepare_ec2_instance.sh
matthieu-d4r: We do, but considering this script lives within such repo, I figured that it should already be cloned 😄

s3torchbenchmarking/utils/prepare_ec2_instance.sh
matthieu-d4r: I recall needing that to build the client, but I believe I was installing it using `-e`, in which case Rust is required... will remove.

s3torchbenchmarking/utils/prepare_ec2_instance.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+#
+# Script to prepare an EC2 instance for PyTorch benchmarks.
+
+set -euo pipefail
```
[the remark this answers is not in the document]
matthieu-d4r: To quote https://ivan.dev/articles/better-shell-scripts:

> - `-e` exits immediately if any simple command fails with a non-zero status. This avoids executing further commands that may have unintended consequences.
> - `-u` treats unset variables as an error when substituting, rather than substituting with an empty string. This helps catch typos in variable names.
> - `-o pipefail` returns a non-zero exit code if any command in the pipeline fails, not just the last one. This helps catch errors in a pipeline.

matthieu-d4r:
> Also, could you explain how we will able to connect results from different EC2 for the same benchmark run?

Well, given that we don't support run "dispatching" / "distribution" yet (i.e., the ability to start multiple benchs, in multiple instances, from a single one), I believe each run will be differentiated with the scenario name.

PR #286, opened 2024-12-17
feat(benchmarks): add Jupyter Notebook for results exploitation

matthieu-d4r:
Add a Jupyter Notebook to fetch and transform benchmark results into Pandas `DataFrame` and CSV file.

--------


IsaevIlya (approved):
I think it is good enough for the start. Later we can add charts/table representation of benchmark data

PR #292, opened 2025-01-03
chore(benchmarks): tidy up benchmark stuffs

matthieu-d4r:
Delete leftover script files. Add minor change to Jupyter Notebook (`"ec2_metadata"` key in results table). Simplify pyproject.toml dependencies list. Change some example parameters in the Hydra config files, add clarifying comments. Delete unit test (not meaningful, for now). Rework the READMEs. Tune the utils/prepare_nvme.sh to work for both Amazon Linux and Ubuntu EC2 instances. Update global .gitignore. Delete utils/prepare_ec2_instance.sh, and add its content to the README. For dataset scenario, add training time measurement around epochs. Minor Python code improvements.

--------

By submitting this pull request, I confirm that my contribution is made under the terms of BSD 3-Clause License and I agree to the terms of the [LICENSE](https://github.com/awslabs/s3-connector-for-pytorch/blob/main/LICENSE).

PR #294, opened 2025-01-08
Ensure path exists, before writing dataset config file

IsaevIlya:
## Description
We have deleted `dataset` folder from `s3torchbenchmarking/conf/`. That caused `datagen` to fail create dataset config file in non existed folder. Added line to create all missing directories in path for dataset config.

--------


IsaevIlya:
https://github.com/awslabs/s3-connector-for-pytorch/pull/292 will handle that issue

PR #296, opened 2025-01-10
Bumping version to 1.3.1

IsaevIlya:
## Description
Bumping version to 1.3.1

--------


s3torchconnector/pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
 
 dependencies = [
     "torch >= 2.0.1, != 2.5.0",
-    "s3torchconnectorclient >= 1.3.0",
```
matthieu-d4r, line 26: nit: do we want to write `==` instead? not blocking

matthieu-d4r (dismissed):

s3torchconnector/pyproject.toml
IsaevIlya: Good question, lets think about that.

matthieu-d4r (approved):

Issue #295, opened 2025-01-08
Saving a distributed checkpoint fails with S3Exception: Service error: The object was not found


dimdi-y:
### s3torchconnector version

I install s3torchconnector via `pip3 install git+https://github.com/awslabs/s3-connector-for-pytorch.git@18c7c12479ced8de78b4a165d7ae9f9b1613af34#subdirectory=s3torchconnector`

### s3torchconnectorclient version

s3torchconnectorclient-1.3.0

### AWS Region

us-east-2

### Describe the running environment

Running on a EC2 instance with Ubuntu 22.04

I install `s3torchconnector` from git because I am using pytorch nightly: `2.6.0.dev20241210+cu126`, which is built from [this commit](https://github.com/pytorch/pytorch/tree/96a8273b6365ba4d205174733ca9a29d991fc850), and the released version requires torch < 2.5.

### What happened?

The connector fails with `S3Exception: Service error: The object was not found`  when trying to store a distributed checkpoint.

The Exception is thrown from `S3FileSystem.rename` method when copying `{self.folder}/.metadata.tmp` file into `{self.folder}/.metadata` in `S3StorageWriter.finish`.

I have checked that the `.metadata.tmp` file is present in the checkpoint folder on s3 before `fs.rename` is called [here](https://github.com/pytorch/pytorch/blob/96a8273b6365ba4d205174733ca9a29d991fc850/torch/distributed/checkpoint/filesystem.py#L595) by adding `assert self.fs.exists(tmp_path)` before it

I have worked around this issue by overriding the `finish` method of the `S3StorageWriter` to write into the final destination directly without creating a temporary file (and therefore without the subsequent rename, which seems to be the culprit).

Here is a short script that reproduces the error (I call the `rename` method directly to simplify the script):
```python

from s3torchconnector.dcp.s3_file_system import S3StorageWriter

S3_REGION = "<INSERT AWS REGION HERE>"
S3_PREFIX = "s3://<INSERT S3 BUCKET HERE>/<INSERT S3 PREFIX HERE>"

writer = S3StorageWriter(
    region=S3_REGION,
    path=S3_PREFIX,
    overwrite=False,
)

final_path = f"{S3_PREFIX}/test.metadata"
tmp_path = f"{final_path}.tmp"

with writer.fs.create_stream(tmp_path, "wb") as metadata_file:
    metadata_file.write(b"test_file_to_rename")

writer.fs.rename(tmp_path, final_path)
```

[...]

PR #297, opened 2025-01-15
Escape special characters in path name for COPY request

IsaevIlya:
## Description
If path contains special characters COPY request will fail. We need to escape special characters for COPY request.

## Related items
https://github.com/awslabs/s3-connector-for-pytorch/issues/295

## Testing
Update integration tests

--------

### DIFF
--- a/s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
+++ b/s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
@@ -127,7 +127,7 @@ class S3FileSystem(FileSystemBase):
         new_path_str = _path_or_str_to_str(new_path)
 
         old_bucket, old_key = parse_s3_uri(old_path_str)
-        escaped_old_key = self._uri_escape_path(old_key)
+        escaped_old_key = self._escape_path(old_key)
         logger.debug("rename: escaped version of the source key: %s", escaped_old_key)
         new_bucket, new_key = parse_s3_uri(new_path_str)
 
@@ -202,14 +202,14 @@ class S3FileSystem(FileSystemBase):
         self._client.delete_object(bucket_name, old_key)
 
     @staticmethod
-    def _uri_escape_path(string):
-        """Return a version of the string with special characters escaped.
+    def _escape_path(string):
+        """URL-encodes path segments while preserving '/' separators using urllib.parse.quote().
 
         Args:
-            string (str): The string to escape.
+            string (str): URL path string to escape
 
         Returns:
-            str: The escaped string.
+            str: Path string with each segment percent-encoded, separators preserved
         """
         if not string:
             return string
--- a/s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
+++ b/s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
@@ -4,6 +4,7 @@
 import io
 import logging
 import os
+import urllib.parse
 from contextlib import contextmanager
 from pathlib import Path
 from typing import Generator, Union, Optional
@@ -126,6 +127,8 @@ class S3FileSystem(FileSystemBase):
         new_path_str = _path_or_str_to_str(new_path)
 
         old_bucket, old_key = parse_s3_uri(old_path_str)
+        escaped_old_key = self._uri_escape_path(old_key)
+        logger.debug("rename: escaped version of the source key: %s", escaped_old_key)
         new_bucket, new_key = parse_s3_uri(new_path_str)
 
         if old_bucket != new_bucket:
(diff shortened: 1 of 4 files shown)


s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
```diff
@@ -126,6 +127,8 @@ def rename(
         new_path_str = _path_or_str_to_str(new_path)
 
         old_bucket, old_key = parse_s3_uri(old_path_str)
+        escaped_old_key = self._uri_escape_path(old_key)
+        logger.debug("rename: escaped version of the source key: %s", escaped_old_key)
```
matthieu-d4r, line 131: i would add the "unescaped" key in the log, so if the conversion failed or whatever, we know the before and after

s3torchconnector/tst/e2e/dcp/test_e2e_s3_file_system.py
```diff
@@ -207,8 +207,9 @@ def test_dcp_load_non_existing_s3_uri(checkpoint_directory):
     print("Test passed: Raised CheckpointException.")
 
 
-def test_successful_rename(checkpoint_directory):
-    src_path = f"{checkpoint_directory.s3_uri}test_rename_src"
+@pytest.mark.parametrize("path", ["test_rename_src", "test_[+rename]!@£$%^&*_(src)"])
```
matthieu-d4r, line 210: the test string does not contain a dot (`.`) 🙂

also, do we want to add a stand-alone test case, comparing the "before" and "after" paths, and making sure they are as expected?

s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
```diff
@@ -198,6 +201,23 @@ def _delete_with_retry(self, bucket_name: str, old_key: str):
         exception too, and logs retries and final error, if any."""
         self._client.delete_object(bucket_name, old_key)
 
+    @staticmethod
+    def _uri_escape_path(string):
+        """Return a version of the string with special characters escaped.
```
matthieu-d4r, line 206: "special characters" is a bit vague in this context: do we have an exhaustive list for which S3 will complain about?

s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
```diff
@@ -198,6 +201,23 @@ def _delete_with_retry(self, bucket_name: str, old_key: str):
         exception too, and logs retries and final error, if any."""
         self._client.delete_object(bucket_name, old_key)
 
+    @staticmethod
+    def _uri_escape_path(string):
```
matthieu-d4r, line 205: nit: since this method is only used on a prefix, it's not really working on a URI; maybe `_escape_path()` is more suited?

s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
```diff
@@ -198,6 +201,23 @@ def _delete_with_retry(self, bucket_name: str, old_key: str):
...
+            string (str): The string to escape.
+
+        Returns:
+            str: The escaped string.
+        """
+        if not string:
+            return string
+        parts = []
+        for part in string.split("/"):
+            parts.append(urllib.parse.quote(part, safe=""))
```
matthieu-d4r, line 218: do we want to leverage `urllib`, even if we're not really dealing with a URL? i know the latter can have some unexpected behaviours... (see my comment below about additional test cases)

use `re` (regex) instead?

matthieu-d4r (changes requested):
do we want to document somewhere in the README(s) about this limitations, so the customers know about it?

also, with this new behaviour, we may be copying files to a totally new path, which may already exist: what do we do in this case?

s3torchconnector/tst/e2e/dcp/test_e2e_s3_file_system.py
IsaevIlya: The issue with a dots in a bucket name. It will be addressed in separate PR. Dot in the path is not escaped, however added it to integration test, just in case.

Added unit test to cover different cases for special characters.

s3torchconnector/tst/unit/dcp/test_s3_file_system.py
```diff
@@ -124,6 +124,34 @@ def raise_s3_exception(bucket, key):
     assert "this was the 3rd time calling it" in caplog.text
 
 
+@pytest.mark.parametrize(
```
matthieu-d4r, line 127: nice!

matthieu-d4r (approved):

PR #300, opened 2025-01-21
Feature/support dots in bucket name

IsaevIlya:
## Description
Consume changes from mountpoint-s3-client. Changes contains support for dots in bucket name for COPY operation that was introduced in CRT.

## Additional context
Internal S3Client now returns `HeadObjectResult` instead of `ObjectInfo`. The only difference is that `HeadObjectResult` doesn't have `key` field.

## Related items
#295

--------

### DIFF
--- a/s3torchconnector/pyproject.toml
+++ b/s3torchconnector/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "s3torchconnector"
-version = "1.3.1"
+version = "1.3.2"
 description = "S3 connector integration for PyTorch"
 requires-python = ">=3.8,<3.13"
 readme = "README.md"
@@ -23,7 +23,7 @@ classifiers = [
 
 dependencies = [
     "torch >= 2.0.1, != 2.5.0",
-    "s3torchconnectorclient == 1.3.1",
+    "s3torchconnectorclient == 1.3.2",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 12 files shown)


s3torchconnector/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "s3torchconnector"
-version = "1.3.1"
```
matthieu-d4r, line 7: why are you bumping s3torchconnector versions? i understand for mountpoint-s3 and CRT, but shouldn't you bump the former through a specific "release" PR/commit?

matthieu-d4r (commented):
LGTM overall, but please add @muddyfish to the review for the Rust part 🙏🏻

s3torchconnector/pyproject.toml
IsaevIlya: There are breaking changes in MountPoint interface, so I need to distinguish versions for testing purpose. With changes s3torchconnector, shouldn't work with previous version of s3torchconnectorclient. Regarding "release", I think it will follow shortly after addressing current issue.

s3torchconnector/tst/e2e/conftest.py
```diff
@@ -46,6 +46,9 @@ def __init__(
     def s3_uri(self):
         return f"s3://{self.bucket}/{self.prefix}"
 
+    def is_standard_storage(self):
+        return self.storage_class != "EXPRESS_ONEZONE"
```
muddyfish, line 50: I'd probably change this to `is_express_storage` and then the following usage to `if not checkpoint_directory.is_express_storage()`

matthieu-d4r (approved):

PR #305, opened 2025-02-03
Update changelog and consume latest MP version

IsaevIlya:
## Description
Update changelog and consume latest mountpoint-s3-client

--------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## TBD
+## v1.3.2 (February 3, 2025)
 
 ### New features
 * Consume mountpoint-s3-client changes with support for dots in bucket name for COPY operation introduced in CRT (#300)
--- a/s3torchconnectorclient/Cargo.lock
+++ b/s3torchconnectorclient/Cargo.lock
@@ -55,9 +55,9 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.83"
+version = "0.1.86"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "721cae7de5c34fbb2acd27e21e6d2cf7b886dce0c27388d46c4e6c47ea4318dd"
+checksum = "644dd749086bf3771a2fbc5f256fdb982d53f011c7d5d560304eafeecebce79d"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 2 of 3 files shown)


muddyfish:
Why is `weights_only` set to False in this PR? I don't see a pytorch update here

IsaevIlya:
> Why is `weights_only` set to False in this PR? I don't see a pytorch update here

PyTorch has made changes following their deprecation of using `weights_only` being default to `False`. For PyTorch part it affects only integration tests for which we have a full control, so there are no risks in using `weights_only=False`. With the new default values `weights_only=True` some of our integration tests would became obsolete, because of restrictions on data types that could be loaded, see [here](https://pytorch.org/docs/stable/notes/serialization.html#weights-only):
```
restricts the unpickler used in torch.load to only executing functions/building classes required for state_dicts of plain torch.Tensors as well as some other primitive types
```
Sample of such error below:
```
Traceback (most recent call last):
    |   File "/home/runner/work/s3-connector-for-pytorch/s3-connector-for-pytorch/s3torchconnector/tst/unit/test_checkpointing.py", line 35, in test_general_checkpointing_saves_python_primitives
    |     _test_save(data, byteorder, use_modern_pytorch_format)
    |   File "/home/runner/work/s3-connector-for-pytorch/s3-connector-for-pytorch/s3torchconnector/tst/unit/test_checkpointing.py", line 147, in _test_save
    |     assert equal(load_with_byteorder(serialised, byteorder), data)
    |   File "/home/runner/work/s3-connector-for-pytorch/s3-connector-for-pytorch/s3torchconnector/tst/unit/_checkpoint_byteorder_patch.py", line 26, in load_with_byteorder
    |     return torch.load(fobj)
    |   File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/torch/serialization.py", line 1470, in load
    |     raise pickle.UnpicklingError(_get_wo_message(str(e))) from None
    | _pickle.UnpicklingError: Weights only load failed. This file can still be loaded, to do so you have two options, do those steps only if you trust the source of the checkpoint. 
    | 	(1) In PyTorch 2.6, we changed the default value of the `weights_only` argument in `torch.load` from `False` to `True`. Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
    | 	(2) Alternatively, to load with `weights_only=True` please check the recommended steps in the following error message.
    | 	WeightsUnpickler error: Unsupported global: GLOBAL decimal.Decimal was not an allowed global by default. Please use `torch.serialization.add_safe_globals([Decimal])` or the `torch.serialization.safe_globals([Decimal])` context manager to allowlist this global if you trust this class/function.
    | 
    | Check the documentation of torch.load to learn more about types accepted by default with weights_only https://pytorch.org/docs/stable/generated/torch.load.html.
    | Falsifying example: test_general_checkpointing_saves_python_primitives(
    |     # The test always failed when commented parts were varied together.
    |     data=Decimal('Infinity'),
    |     byteorder='little',  # or any other generated value
    |     use_modern_pytorch_format=True,  # or any other generated value
    | )
    | Explanation:
    |     These lines were always and only run by failing examples:
    |         /opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/torch/_weights_only_unpickler.py:334
    ```

PR #306, opened 2025-02-04
Feature/handle pytorch load changes

IsaevIlya:
## Description
In PyTorch 2.6, `torch.load()` now requires explicit `weights_only` parameter. When defaulting to `True`, 
it restricts supported data types during loading (see [serialization docs](https://pytorch.org/docs/stable/notes/serialization.html#weights-only)).
Set `weights_only=False` in `S3LightningCheckpoint` to match PyTorch Lightning's `CheckpointIO` implementation, maintain backwards compatibility.

--------

### DIFF
--- a/s3torchconnector/tst/unit/_hypothesis_python_primitives.py
+++ b/s3torchconnector/tst/unit/_hypothesis_python_primitives.py
@@ -3,47 +3,32 @@
 
 from hypothesis.strategies import (
     integers,
-    binary,
-    none,
     characters,
-    complex_numbers,
     floats,
     booleans,
-    decimals,
-    fractions,
     deferred,
-    frozensets,
     tuples,
     dictionaries,
     lists,
-    uuids,
-    sets,
     text,
 )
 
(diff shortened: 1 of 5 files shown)


s3torchconnector/tst/unit/_hypothesis_python_primitives.py
```diff
@@ -3,47 +3,30 @@
...
     lists,
-    uuids,
-    sets,
     text,
 )
 
 scalars = (
-    none()
-    | booleans()
+    booleans()
```
muddyfish, line 17: Depending on what we care about for 'python primitives' this is not a complete list :)

muddyfish (approved):

s3torchconnector/tst/unit/_hypothesis_python_primitives.py
IsaevIlya: I agree, that it is primitives that now supported by torch.load

PR #313, opened 2025-02-26
Add another DCP benchmark that is using FSDP with llama v2 model

IsaevIlya:
## Description
Our existing DCP benchmark uses DDP, which limits our ability to run bigger models on GPU. With introducing support of FSDP we will be able to run DCP bebchmark for bigger models on GPU without falling back to CPU. 

## Additional context
Existed DCP benchmark was renamed to dcp_ddp to preserve it.

--------


s3torchbenchmarking/conf/dcp_fsdp.yaml
```diff
@@ -0,0 +1,38 @@
...
+      # Checkpoint storage location (valid options: "disk", "s3").
+      +checkpoint.storage: disk, s3
+      # Sharding strategy (valid options: "full", "hybrid").
+      +checkpoint.sharding_strategy: full
+      # Controls whether files are forcibly synced to disk (only relevant for "disk" storage).
+      # NOTE: We disabled this option to improve performance since FSDP checkpointing with
+      # forced syncing (maximum durability) was significantly slower than storage throughput.
+      # This setting has no effect when using "s3" storage.
+      +checkpoint.sync_files: false
+
```
rajdchak, line 38: Instead of having two different file dcp_fsdp.yaml and dcp_ddb.yaml, is there a way to combine these two into one dcp.yaml file?

IsaevIlya: While it's technically possible to merge the FSDP and DDP configuration files into a single file, doing so would create several challenges:
- Model Compatibility: FSDP and DDP benchmarks support different sets of models. Combining configurations could lead to users accidentally selecting models that aren't compatible with their chosen distribution strategy.
- Configuration Clarity: A merged configuration would include parameters that are only relevant to one distribution strategy but not the other. This could make the configuration file more confusing and increase the likelihood of setup errors.
My main idea was to keep it simple for usage and prevent probability of misconfiguration.

s3torchbenchmarking/src/s3torchbenchmarking/dcp_fsdp/benchmark.py
```diff
@@ -0,0 +1,211 @@
...
+from s3torchbenchmarking.models import get_benchmark_model
+from s3torchconnector.dcp import S3StorageWriter
+
+Timestamps = Tuple[float, float]
+logger = logging.getLogger(__name__)
+
+
+# TODO: add Structured Config (https://hydra.cc/docs/tutorials/structured_config/intro/)
+@hydra.main(version_base=None)
+def run_benchmark(cfg: DictConfig) -> dict:
```
rajdchak, line 42: There is a lot of common code between dcp_ddp and dcp_fsdp benchmark classes, can we move all the common methods/code to a separate utils class?

s3torchbenchmarking/src/s3torchbenchmarking/dcp_fsdp/benchmark.py
vaibhav5140, line 1: There are some common functions shared between dcp_ddp/benchmark.py and dcp_fsdp/benchmark.py (like get_writer(), setup(), and run_benchmark()).
Can we create a base class that both implementations could inherit from?

s3torchbenchmarking/src/s3torchbenchmarking/dcp_fsdp/benchmark.py
vaibhav5140: +1

s3torchbenchmarking/src/s3torchbenchmarking/dcp_fsdp/benchmark.py
vaibhav5140: Also, the 'dcp' prefix is used in multiple places (dcp_fsdp, dcp_ddp). Should we consider a structure like:
  |--dcp/  # All dcp-related code in one place
    |--base/  # Shared/common code
    |---ddp_impl/  # ddp-specific implementation
    |---fsdp_impl/  # fsdp-specific implementation

IsaevIlya: Changes in folder structure will require to modify runner script to support such folder ctructure, which will make runner more complicated without bringing any new value.

s3torchbenchmarking/conf/dcp_fsdp.yaml
rajdchak: okay

s3torchbenchmarking/src/s3torchbenchmarking/benchmark_utils.py
```diff
@@ -116,7 +117,8 @@ def stop(self):
 
 def build_random_suffix() -> str:
     """Build a random suffix for use in filepaths or S3 URIs."""
-    return "".join(random.choices(string.ascii_letters, k=7))
+    prefix = datetime.now().strftime("%Y-%m-%d-%H-%M")
```
rajdchak, line 120: Nit: can name this to date instead of prefix.

PR #320, opened 2025-03-25
New work around for fork with multithread from CRT

TingDaoK:
## Description
Previously, since the underlying CRT client has multiple background threads, there are couple hacks have done to workaround the fork issue. **Notes: overall, fork with multi-threading is not encouraged and not well supported.**
1. check the PID to make sure every process creates its own s3 client
2. If the client was not created by the current process, leave the client in memory to avoid the clean up steps try to invoke background threads already gone.

Well, it works **mostly**.
CRT has some global state related to threads (There may be more than what I listed, I just listed what I have found so far)
1. https://github.com/awslabs/aws-c-common/pull/1181 for more details
2. https://github.com/awslabs/aws-c-common/blob/main/source/thread_shared.c#L16, there is a global mutex a background thread can grab when it finises running. So, if the fork happens when one background threads holding the lock, that lock will be a dead lock forever for the subprocess.

Given the findings above, a _better_ workaround is to drop the CRT S3 client and then wait for CRT background threads to join, which will at least get around the two findings listed above. Printout warnings for user about the potential risk.

THIS IS JUST A PROVE OF CONCEPT, feel free to take over the branch 🙇‍♂️ 

## Additional context

## Testing

--------

### DIFF
--- a/s3torchconnector/src/s3torchconnector/_s3client/_s3client.py
+++ b/s3torchconnector/src/s3torchconnector/_s3client/_s3client.py
@@ -3,6 +3,7 @@
 
 import logging
 import os
+import gc
 import threading
 from functools import partial
 from typing import Optional, Any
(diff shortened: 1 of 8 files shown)


s3torchconnector/src/s3torchconnector/_s3client/_s3client.py
```diff
@@ -48,25 +75,22 @@ def __init__(
...
-                if self._client_pid is None or self._client_pid != os.getpid():
-                    # `MountpointS3Client` does not survive forking, so re-create it if the PID has changed.
-                    self._real_client = self._client_builder()
-                    self._client_pid = os.getpid()
-        assert self._real_client is not None
-        return self._real_client
+                CRT_S3_CLIENT = self._client_builder()
+
+        assert CRT_S3_CLIENT is not None
+        return CRT_S3_CLIENT
```
IsaevIlya, line 93: Could you please incorporate your changes into existing logic?
While `_before_fork_handler` cleans up `CRT_S3_CLIENT` for 'fork' method,  the global `CRT_S3_CLIENT` becomes invalid when using 'spawn' or 'forkserver' since `MountpointS3Client` won't survive serialization/de-serialization. This will lead to errors when attempting to use the client in child processes.

s3torchconnector/tst/e2e/test_mountpoint_client_parallel_access.py
```diff
@@ -10,12 +10,17 @@
 class S3ClientWithoutLock(S3Client):
     @property
     def _client(self) -> MountpointS3Client:
-        if self._client_pid is None or self._client_pid != os.getpid():
+        global NATIVE_S3_CLIENT
+        if (
+            self._client_pid is None
+            or self._client_pid != os.getpid()
+            or NATIVE_S3_CLIENT is None
```
IsaevIlya, line 17: Let's remove the third condition `or NATIVE_S3_CLIENT is None` as it preventing negative test case to happen. Also, lets declare `NATIVE_S3_CLIENT = None` before class definition to be sure that we are starting with non existing client.

PR #352, opened 2025-07-24
Experiment/dcp save ordering

IsaevIlya:
## Description
Improved Performance for Partial Checkpoint Loading

### Background
By default, PyTorch sorts checkpoint data based on size, which distributes tensors/weights randomly across checkpoint shards. While this approach works well with local storage, it can impact performance when working with cloud storage. Currently, PyTorch's checkpoint loading process doesn't follow the same order used during saving, resulting in non-sequential file access patterns.

#### 1. Sequential Read Optimization
- Modified the ordering of items in LocalPlan based on their actual offset in checkpoint shards
- Ensures sequential reading of data, improving I/O efficiency

#### 2. Custom Sorting for Partial Loading
- Added ability to provide a custom sorting key function during checkpoint saving
- Allows users to group specific data at the beginning of checkpoints
- Optimizes partial loading scenarios by reducing the amount of data that needs to be read

### Usage Example
If you need to load only model layers that start with "model.model.layers.1", you can group these tensors at the beginning of checkpoint shards:

```python
def sort_key_func(item):
    return not item.index.fqn.startswith("model.model.layers.1"), item.index.fqn

def save_checkpoint(state_dict, region, uri):
    writer = S3StorageWriter(region, uri, sort_key=sort_key_func)
    dcp.save(state_dict, storage_writer=storage_writer)

``` 

#### Results

The resulting checkpoint shard layout will prioritize tensors starting with "model.model.layers.1":
```
File name: __0_0.distcp
        key: model.model.layers.10.self_attn.o_proj.weight     ,        offset: 0
        key: model.model.layers.11.mlp.down_proj.weight        ,        offset: 67110441
        key: model.model.layers.11.self_attn.k_proj.weight     ,        offset: 247467090
        key: model.model.layers.13.mlp.gate_proj.weight        ,        offset: 314577531
        key: model.model.layers.15.input_layernorm.weight      ,        offset: 494934180
        key: model.model.layers.15.post_attention_layernorm.weight,     offset: 494952141
        ...
        key: model.lm_head.weight                              ,        offset: 1124125164
        key: model.model.embed_tokens.weight                   ,        offset: 1950404629
        key: model.model.layers.21.mlp.gate_proj.weight        ,        offset: 2776684094
        ....
```

--------

### DIFF
--- a/s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
+++ b/s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
@@ -7,7 +7,7 @@ import os
 import urllib.parse
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Generator, Union, Optional
+from typing import Generator, Union, Optional, Callable, Any
 from typing import List
 
 from s3torchconnectorclient._mountpoint_s3_client import S3Exception
@@ -252,7 +252,7 @@ class S3FileSystem(FileSystemBase):
         return "/".join(parts)
 
 
-from torch.distributed.checkpoint.planner import SavePlan
+from torch.distributed.checkpoint.planner import SavePlan, LoadPlan
 import dataclasses
 from dataclasses import dataclass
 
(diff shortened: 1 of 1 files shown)


jet-tong (commented):
Tested the PR with "model.model.layers.1" sort key function, and can confirm that these changes speed up partial checkpoint loading.

jet-tong:
Closing since we adopted a new approach in https://github.com/awslabs/s3-connector-for-pytorch/pull/378 to address general use case of DCP partial loading. 

For users: Note this patch could still speed up partial loading by grouping required tensors together at the beginning of the checkpoint files, but monkey-patching `torch.distributed.checkpoint.filesystem` `_split_by_size_and_type` function makes it harder to maintain long term.

PR #356, opened 2025-08-20
Feature/multi gpu dcp benchmarks

MSH4R1F:
## Description
Extended S3 Data Loading Benchmarks to support multi-GPU distributed traininng

## Additional context
Our current benchmarks currently only support a single GPU training. If you used multiple GPUs this would result in all GPUs except for 1 to be under-utilized.

## Changes Made
• **Multi-GPU Support**: Added distributed data parallel (DDP) training using torch.multiprocessing.spawn
• **Dataset Sharding**: 

- S3IterableDataset: We use enable_sharding=True for automatic data distribution as well as a wrapper around dataloader when enumerating to make sure that that each shard has an equal batch size and so we can cap steps to a global minimum
-  S3MapDataset: Uses DistributedSampler for proper data partitioning

• **Resource Monitoring**: Extended to track all GPU utilization instead of single GPU
• Environment optimizations (OMP_NUM_THREADS=1, NCCL_P2P_DISABLE=1)
• Added Dataset validation make sure no duplicate processing

## Performance Results
**S3IterableDataset (100k images, 134MB shards):**
• **Multi-GPU (4 GPUs)**: 218s/epoch, 82.6% avg utilization across all GPUs
• **Single-GPU**: 750s/epoch, 86% utilization on GPU 0 only
• **Improvement**: 3.5x throughput increase

**S3MapDataset (ViT model):**
• **Multi-GPU (4 GPUs)**: 231s training, ~78% utilization across all GPUs, 105 MiB/s
• **Single-GPU**: 817s training, 81% utilization on GPU 0 only, 122 MiB/s  
• **Improvement**: 3.5x throughput increase

### Logic in Capped Loader
One issue of running multi-GPU training is that the data sharded between the GPUs must be the same size. Where this isn't possible, one solution to ensure even splitting is to trim the dataset so each GPUs contains the same number of batches. If this doesn't happen then there occurs a hang within the GPUs causing the training to fail.
With S3MapDataset, this is easy to do as we already know the size of the dataset and therefore we can use `len()` to discover the size of the dataset and therefore can know the maximum number of datapoints each GPU can train on whilst being evenly split.

With S3IterableDataset, this is harder to do as we don't know the size of the dataset and therefore can't know how to evenly split between the GPUs. My solution to this was to yield a batch from either GPU simultaneously whilst it is possible and if it isn't possible to yield another batch we stop our training at that moment.

We ensure quick inter-process communication using `dist.all_gather_objects` and therefore can early-stop when needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * Fix SequentialS3Reader to stay at EOF when seeking beyond object size (#362)
 
 ### Other changes
+* Extended benchmarks to support multi-GPU Dataloading
 * Add seekable() method in S3Reader to eliminate tensor copies during DCP loading (#359)
 * Add load ordering optimization to S3StorageReader for sequential access patterns (#372)
 * Add os, arch, and PyTorch version to user agent string (#397)
@@ -31,9 +32,6 @@
 
 ### Breaking changes
 * No breaking changes, but DCPOptimizedS3Reader as the new default reader for `S3StorageReader` might lead to behavioral changes. See [DCPOptimizedS3Reader Errors](https://github.com/awslabs/s3-connector-for-pytorch/blob/main/docs/TROUBLESHOOTING.md#dcpoptimizeds3reader-errors) for more details.
-
-### Other changes
-* Extended benchmarks to support multi-GPU Dataloading
 
 ## v1.4.3 (July 25, 2025)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@
 ### Breaking changes
 * No breaking changes, but DCPOptimizedS3Reader as the new default reader for `S3StorageReader` might lead to behavioral changes. See [DCPOptimizedS3Reader Errors](https://github.com/awslabs/s3-connector-for-pytorch/blob/main/docs/TROUBLESHOOTING.md#dcpoptimizeds3reader-errors) for more details.
 
+### Other changes
+* Extended benchmarks to support multi-GPU Dataloading
+
 ## v1.4.3 (July 25, 2025)
 
 ### New features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Other changes
 * Bump PyO3 from 0.27.2 to 0.29.0
+* Extended benchmarks to support multi-GPU Dataloading (#356)
 
 ### Breaking changes
 
@@ -28,7 +29,6 @@
 * Add benchmark to run DCP Loading Workloads (#357)
 * Add thread_count parameter to S3StorageWriter (#370)
 * Add macOS x86_64 and Python 3.8 deprecation warnings (#400)
-* Extended benchmarks to support multi-GPU Dataloading (#356)
 
 ### Breaking changes
 * No breaking changes, but DCPOptimizedS3Reader as the new default reader for `S3StorageReader` might lead to behavioral changes. See [DCPOptimizedS3Reader Errors](https://github.com/awslabs/s3-connector-for-pytorch/blob/main/docs/TROUBLESHOOTING.md#dcpoptimizeds3reader-errors) for more details.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,13 +22,13 @@
 * Fix SequentialS3Reader to stay at EOF when seeking beyond object size (#362)
 
 ### Other changes
-* Extended benchmarks to support multi-GPU Dataloading
 * Add seekable() method in S3Reader to eliminate tensor copies during DCP loading (#359)
 * Add load ordering optimization to S3StorageReader for sequential access patterns (#372)
 * Add os, arch, and PyTorch version to user agent string (#397)
 * Add benchmark to run DCP Loading Workloads (#357)
 * Add thread_count parameter to S3StorageWriter (#370)
 * Add macOS x86_64 and Python 3.8 deprecation warnings (#400)
+* Extended benchmarks to support multi-GPU Dataloading (#356)
 
 ### Breaking changes
 * No breaking changes, but DCPOptimizedS3Reader as the new default reader for `S3StorageReader` might lead to behavioral changes. See [DCPOptimizedS3Reader Errors](https://github.com/awslabs/s3-connector-for-pytorch/blob/main/docs/TROUBLESHOOTING.md#dcpoptimizeds3reader-errors) for more details.
--- a/s3torchbenchmarking/README.md
+++ b/s3torchbenchmarking/README.md
@@ -93,6 +93,27 @@ Here are some sample dataset configurations that we ran our benchmarks against:
 s3torch-datagen -n 100k --shard-size 128MiB --s3-bucket my-bucket --region us-east-1
 ```
 
+## Multi-GPU dataset benchmarks
+
+Setting `num_gpus > 1` in `conf/dataset.yaml` enables multi-GPU benchmarking via DDP (DistributedDataParallel).
+This measures data loading throughput under realistic multi-GPU training conditions where multiple processes
+pull data from S3 simultaneously.
+
+**Behavior when `num_gpus > 1`:**
+
+- One process is spawned per GPU via `mp.spawn`
+- Each process runs the full training loop independently on its assigned GPU
+- A `capped_loader` ensures all ranks process the same number of batches to prevent DDP hangs:
+  - Map-style datasets (S3MapDataset): uses `DistributedSampler` to shard indices across ranks; `capped_loader` caps iteration to `min(len)` across all ranks
+  - Iterable datasets (S3IterableDataset): enables `enable_sharding` to split objects across ranks; `capped_loader` synchronizes per-batch and stops all ranks when any rank exhausts its data
+- `drop_last` is forced to `true` (DDP requires equal batch sizes across ranks)
+- Some samples may be skipped due to capping/dropping; check `volume` in metrics for actual count
+
+**Constraints:**
+
+- `model=entitlement` always runs single-GPU (I/O-only benchmark, no gradient sync needed)
+- Only `s3iterabledataset` and `s3mapdataset` support multi-GPU sharding; other backends run unsharded
+
 ## Running the benchmarks
 
 You can run the different benchmarks by editing their corresponding config files, then running one of those shell
--- a/s3torchbenchmarking/README.md
+++ b/s3torchbenchmarking/README.md
@@ -202,8 +202,8 @@ pull data from S3 simultaneously.
 - A `capped_loader` ensures all ranks process the same number of batches to prevent DDP hangs:
   - Map-style datasets (S3MapDataset): uses `DistributedSampler` to shard indices across ranks; `capped_loader` caps iteration to `min(len)` across all ranks
   - Iterable datasets (S3IterableDataset): enables `enable_sharding` to split objects across ranks; `capped_loader` synchronizes per-batch and stops all ranks when any rank exhausts its data
-- `drop_last` is forced to `true` (DDP requires equal batch sizes across ranks)
-- Some samples may be skipped due to capping/dropping; check `volume` in metrics for actual count
+- `drop_last` defaults to `true` under multi-GPU so a short final batch doesn't skew per-rank step counts. Setting it to `false` is safe — `capped_loader` already keeps ranks in step, so it won't hang.
+- Some samples may be skipped due to capping/dropping; check `volume` in metrics for the actual count
 
 **Constraints:**
 
--- a/s3torchbenchmarking/README.md
+++ b/s3torchbenchmarking/README.md
@@ -93,27 +93,6 @@ Here are some sample dataset configurations that we ran our benchmarks against:
 s3torch-datagen -n 100k --shard-size 128MiB --s3-bucket my-bucket --region us-east-1
 ```
 
-## Multi-GPU dataset benchmarks
-
-Setting `num_gpus > 1` in `conf/dataset.yaml` enables multi-GPU benchmarking via DDP (DistributedDataParallel).
-This measures data loading throughput under realistic multi-GPU training conditions where multiple processes
-pull data from S3 simultaneously.
-
-**Behavior when `num_gpus > 1`:**
-
-- One process is spawned per GPU via `mp.spawn`
-- Each process runs the full training loop independently on its assigned GPU
-- A `capped_loader` ensures all ranks process the same number of batches to prevent DDP hangs:
-  - Map-style datasets (S3MapDataset): uses `DistributedSampler` to shard indices across ranks; `capped_loader` caps iteration to `min(len)` across all ranks
-  - Iterable datasets (S3IterableDataset): enables `enable_sharding` to split objects across ranks; `capped_loader` synchronizes per-batch and stops all ranks when any rank exhausts its data
-- `drop_last` is forced to `true` (DDP requires equal batch sizes across ranks)
-- Some samples may be skipped due to capping/dropping; check `volume` in metrics for actual count
-
-**Constraints:**
-
-- `model=entitlement` always runs single-GPU (I/O-only benchmark, no gradient sync needed)
-- Only `s3iterabledataset` and `s3mapdataset` support multi-GPU sharding; other backends run unsharded
-
 ## Running the benchmarks
 
 You can run the different benchmarks by editing their corresponding config files, then running one of those shell
@@ -209,3 +188,24 @@ will also be written to the specified table.
 [credentials]: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html
 
 [hydra-overrides]: https://hydra.cc/docs/advanced/override_grammar/basic/
+
+## Multi-GPU dataset benchmarks
+
+Setting `num_gpus > 1` in `conf/dataset.yaml` enables multi-GPU benchmarking via DDP ([DistributedDataParallel](https://docs.pytorch.org/tutorials/intermediate/ddp_tutorial.html)).
+This measures data loading throughput under realistic multi-GPU training conditions where multiple processes
+pull data from S3 simultaneously.
+
+**Behavior when `num_gpus > 1`:**
+
+- One process is spawned per GPU via `mp.spawn`
+- Each process runs the full training loop independently on its assigned GPU
+- A `capped_loader` ensures all ranks process the same number of batches to prevent DDP hangs:
+  - Map-style datasets (S3MapDataset): uses `DistributedSampler` to shard indices across ranks; `capped_loader` caps iteration to `min(len)` across all ranks
+  - Iterable datasets (S3IterableDataset): enables `enable_sharding` to split objects across ranks; `capped_loader` synchronizes per-batch and stops all ranks when any rank exhausts its data
+- `drop_last` is forced to `true` (DDP requires equal batch sizes across ranks)
+- Some samples may be skipped due to capping/dropping; check `volume` in metrics for actual count
+
+**Constraints:**
+
+- `model=entitlement` always runs single-GPU (I/O-only benchmark, no gradient sync needed)
+- Only `s3iterabledataset` and `s3mapdataset` support multi-GPU sharding; other backends run unsharded
--- a/s3torchbenchmarking/conf/dataset.yaml
+++ b/s3torchbenchmarking/conf/dataset.yaml
@@ -40,3 +40,4 @@ hydra:
       dataloader.s3reader.type: sequential
       # buffer_size (bytes): only used with range_based s3reader
       dataloader.s3reader.buffer_size: 8*1024*1024
+      +num_gpus: 1
--- a/s3torchbenchmarking/conf/dataset.yaml
+++ b/s3torchbenchmarking/conf/dataset.yaml
@@ -40,17 +40,12 @@ hydra:
       dataloader.s3reader.type: sequential
       # buffer_size (bytes): only used with range_based s3reader
       dataloader.s3reader.buffer_size: 8*1024*1024
-      # Number of GPUs to use for training.
-      # When num_gpus > 1:
-      #   - Uses DDP (DistributedDataParallel) via mp.spawn
-      #   - S3IterableDataset enables sharding across ranks
-      #   - S3MapDataset uses DistributedSampler
-      #   - capped_loader ensures all ranks process the same number of batches
-      # When num_gpus = 1 (or model=entitlement):
-      #   - Single process, no DDP overhead
+      # Number of GPUs to use (default: 1 = single process, no DDP).
+      # Setting num_gpus > 1 enables DDP via mp.spawn. See README for details.
+      # Note: model=entitlement always forces num_gpus=1 (I/O-only, no training).
       +num_gpus: 1
-      # Whether to drop the last incomplete batch from each worker.
-      # "auto" (default): True for multi-GPU (required for DDP synchronization),
-      #                    False for single-GPU (all data is processed).
-      # Can be overridden to "true" or "false" explicitly.
+      # Whether to drop the last incomplete batch.
+      # "auto" (default): True when num_gpus > 1, False otherwise.
+      # Multi-GPU requires drop_last=true for DDP synchronization; setting it to
+      # false with num_gpus > 1 will cause training to hang.
       +drop_last: auto
--- a/s3torchbenchmarking/conf/dataset.yaml
+++ b/s3torchbenchmarking/conf/dataset.yaml
@@ -40,4 +40,17 @@ hydra:
       dataloader.s3reader.type: sequential
       # buffer_size (bytes): only used with range_based s3reader
       dataloader.s3reader.buffer_size: 8*1024*1024
+      # Number of GPUs to use for training.
+      # When num_gpus > 1:
+      #   - Uses DDP (DistributedDataParallel) via mp.spawn
+      #   - S3IterableDataset enables sharding across ranks
+      #   - S3MapDataset uses DistributedSampler
+      #   - capped_loader ensures all ranks process the same number of batches
+      # When num_gpus = 1 (or model=entitlement):
+      #   - Single process, no DDP overhead
       +num_gpus: 1
+      # Whether to drop the last incomplete batch from each worker.
+      # "auto" (default): True for multi-GPU (required for DDP synchronization),
+      #                    False for single-GPU (all data is processed).
+      # Can be overridden to "true" or "false" explicitly.
+      +drop_last: auto
--- a/s3torchbenchmarking/conf/dataset.yaml
+++ b/s3torchbenchmarking/conf/dataset.yaml
@@ -44,11 +44,7 @@ hydra:
       # Setting num_gpus > 1 enables DDP via mp.spawn. See README for details.
       # Note: model=entitlement always forces num_gpus=1 (I/O-only, no training).
       +num_gpus: 1
-      # Whether each rank drops its final batch when it has fewer than batch_size samples.
-      # "auto" (default): True when num_gpus > 1, False otherwise.
-      # This won't hang if set to false under multi-GPU: capped_loader already
-      # equalises the step count across ranks (map: min(len); iterable: stop when
-      # any rank exhausts), and that step-count sync is what prevents DDP hangs.
-      # drop_last defaults to true for multi-GPU only to keep per-rank step counts
-      # even so throughput/samples-per-epoch stay comparable across ranks.
+      # Whether each rank drops its final short batch (fewer than batch_size samples).
+      # "auto" (default): true under multi-GPU, false otherwise. See README for when
+      # to override and why false is still safe under multi-GPU.
       +drop_last: auto
--- a/s3torchbenchmarking/conf/dataset.yaml
+++ b/s3torchbenchmarking/conf/dataset.yaml
@@ -44,8 +44,11 @@ hydra:
       # Setting num_gpus > 1 enables DDP via mp.spawn. See README for details.
       # Note: model=entitlement always forces num_gpus=1 (I/O-only, no training).
       +num_gpus: 1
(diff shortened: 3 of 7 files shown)


jet-tong (commented):
Great PR overall, most comements are nits, just a few important things to address:
- Small errors (missing brackets, utils vs mem, etc)
- capped loader: should address double loading issue (if I'm seeing it right), and emit details on how much data we skipped loading / training
- Maybe a way to determine how many GPUs to use would be good (or even 1 GPU vs all GPU). Would allow us to run our previous benchmark with 1 GPU and compare them.

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -5,35 +5,129 @@
...
+
+
+def run_benchmark_experiment(config: DictConfig):
+
+    rank = dist.get_rank() if dist.is_initialized() else 0
     model = make_model(config)
 
     fully_qualified_uri = (
-        "s3://" + config.s3.bucket.strip("/") + "/" + config.dataset.strip("/")
+        "s3://" + config.s3.bucket.strip("/") + "/" + config.dataset.strip("/") + "/"
```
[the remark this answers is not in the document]
MSH4R1F: Yepp, thanks!

s3torchbenchmarking/src/s3torchbenchmarking/models.py
```diff
@@ -154,12 +169,25 @@ def train(self, dataloader: DataLoader, epochs: int) -> ExperimentResult:
...
-                    result = self.train_batch(batch_idx, data, target)
-                    num_samples += len(data)
-                    if result:
-                        checkpoint_times.append(result)
+                batch_count = 0
+                try:
+                    for batch_idx, (data, target) in enumerate(
+                        self.capped_loader(dataloader)
+                    ):
+                        logger.debug("Batch #%i", batch_idx)
```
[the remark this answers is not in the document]
MSH4R1F: Yep will allow logs to be shown when completing an epoch, i think this is useful for people to keep track of the progress of trainings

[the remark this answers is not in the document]
MSH4R1F: 🚀

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -5,35 +5,129 @@
...
 )
+
 from s3torchconnector import S3MapDataset, S3Reader, S3IterableDataset
 from s3torchconnector.s3reader import S3ReaderConstructor, S3ReaderConstructorProtocol
 from s3torchconnector._s3dataset_common import parse_s3_uri  # type: ignore
+import torch
+import logging
+import torch.multiprocessing as mp
+import json
+import tempfile
```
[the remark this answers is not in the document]
MSH4R1F: 🚀

s3torchbenchmarking/src/s3torchbenchmarking/models.py
```diff
@@ -153,13 +196,27 @@ def train(self, dataloader: DataLoader, epochs: int) -> ExperimentResult:
...
             for epoch in range(epochs):
                 begin_epoch = time.perf_counter()
-                logger.info("Epoch #%i/%i", epoch, epochs - 1)
-                for batch_idx, (data, target) in enumerate(dataloader):
-                    logger.debug("Batch #%i", batch_idx)
-                    result = self.train_batch(batch_idx, data, target)
-                    num_samples += len(data)
-                    if result:
-                        checkpoint_times.append(result)
+                if dist.get_rank() == 0:
```
jet-tong, line 199: This should be `if not dist.is_initialized() or dist.get_rank() == 0:` so it works for non-distributed training.

s3torchbenchmarking/src/s3torchbenchmarking/models.py
```diff
@@ -142,6 +143,48 @@ def load_sample(self, sample: Union[S3Reader, Tuple[str, IOBase]]):
     def train_batch(self, batch_idx: int, data, target) -> Optional[Any]:
         raise NotImplementedError
 
+    def capped_loader(self, loader):
```
tadiwa-aizen, line 146: Would be helpful to add some logging in here for multi-GPU debugging — things like which rank we're on, whether we're using the map-style `len()` cap or iterable-style per-batch sync, what `min_steps` ended up being, and when we stop (with the flags).

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -225,14 +381,25 @@ def create_fsspec_dataset(
...
 
-def make_dataloader(dataset: Dataset, num_workers: int, batch_size: int):
+def make_dataloader(dataset: Dataset, num_workers: int, batch_size: int, sampler=None):
     return DataLoader(
         dataset=dataset,
         batch_size=batch_size,
+        sampler=sampler,
         shuffle=False,
-        drop_last=False,
+        drop_last=True,
```
tadiwa-aizen: So `drop_last=True` makes sense for multi-GPU DDP (prevents deadlocks when ranks have uneven batches), but it's also kicking in for single-GPU/entitlement runs where it's not needed. It ends up silently dropping the last partial batch, which inflates throughput by ~5% and means fewer samples are actually processed (e.g. 9,472 vs 10,000 per epoch).

 I ran a normalized benchmark with the same DataLoader settings on both branches and got 0% throughput difference — so the structural changes here are clean, it's just this config that skews things.

 What about making this a configurable `drop_last` param that defaults to `"auto"`?
 - `"auto"` → True when DDP is active, False for single-GPU
- Users can still override to `"true"` or `"false"` from the yaml if they wan

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -42,15 +150,32 @@ def run_experiment(config: DictConfig) -> dict:
...
+            # Recreate optimizer AFTER DDP wrapping
+            model._optimizer = None  # Clear cached optimizer
+            model.optimizer = torch.optim.Adam(model.model.parameters(), lr=0.001)
 
     result: ExperimentResult = model.train(dataloader, config.epochs)
 
     metrics = {
         "throughput_mibs": result["volume"] / result["training_duration_s"],
         "training_duration_s": result["training_duration_s"],
+        "volume_mibs": result[
```
tadiwa-aizen: Could we also add a `"samples_per_epoch": result["volume"] / config.epochs` field here? That way when we look at DynamoDB results it's immediately obvious if `drop_last` reduced the per-epoch sample count

s3torchbenchmarking/conf/dataset.yaml
```diff
@@ -39,4 +39,5 @@ hydra:
       # s3reader type: sequential or range_based
       dataloader.s3reader.type: sequential
       # buffer_size (bytes): only used with range_based s3reader
-      dataloader.s3reader.buffer_size: 8*1024*1024
\ No newline at end of file
+      dataloader.s3reader.buffer_size: 8*1024*1024
+      +num_gpus: 1
```
tadiwa-aizen: Can we add some comments here explaining what `num_gpus` actually controls? Something like:
When >1: uses DDP via `mp.spawn`, `S3IterableDataset` enables sharding, `S3MapDataset` uses `DistributedSampler`, `capped_loader` kicks in
- When =1 (or model=entitlement): single process, no DDP overhead

 Also if we go with the `drop_last` fix, we'd want to add `+drop_last: auto` as a sweep param here too.

tadiwa-aizen (changes requested):
Firstly, thank you for all the work on this. @MSH4R1F  We are happy to take over the PR and commit the changes necessary to get this merged.

MSH4R1F:
Thanks tadiwa, nice catching up

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * Fix SequentialS3Reader to stay at EOF when seeking beyond object size (#362)
 
 ### Other changes
+* Extended benchmarks to support multi-GPU Dataloading
```
jet-tong: Would appreciate placing this at the end (s3torchbenchmarking updates after s3torchconnector updates), and including the PR number.

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -37,7 +38,12 @@ def init_distributed(rank=0, world_size=1):
     """Initialize DDP Process group"""
     os.environ["MASTER_ADDR"] = "localhost"
     os.environ["MASTER_PORT"] = "12355"
-    dist.init_process_group(backend="nccl", rank=rank, world_size=world_size)
+    dist.init_process_group(
+        backend="nccl",
+        rank=rank,
+        world_size=world_size,
+        timeout=datetime.timedelta(minutes=5),
```
jet-tong: May I ask why the timeout was added? 

I understand timeouts are nice to have in case of failures, but I don't think we use it in other `dist.init_process_group` calls in this repo so would lean towards not adding too many other changes.

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -80,8 +86,11 @@ def run_experiment(config: DictConfig) -> dict:
             )
             # Read results from file
             if os.path.exists(results_file):
-                with open(results_file, "r") as f:
-                    return json.load(f)
+                try:
+                    with open(results_file, "r") as f:
+                        return json.load(f)
+                except (json.JSONDecodeError, IOError):
+                    return {"metrics": "DDP training completed - results file corrupted"}
```
jet-tong: May I ask why JSONDecodeError was added? This change swallows the JSONDecodeError error - would be good practice to raise the error when something went wrong

Rather than returning {"metrics": "<some bad strings"} (here and in else statement below), perhaps it's better practice in general to raise an error instead of silent failures to help debugging.

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -148,17 +157,19 @@ def run_benchmark_experiment(config: DictConfig):
         region=config.s3.region,
         load_sample=model.load_sample,
     )
+    drop_last_cfg = config.get("drop_last", "auto")
```
jet-tong: Since drop_last is only used in 1 place anyways, would be good to follow existing pattern to just fetch the config param in `make_dataloader`.

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -23,34 +31,161 @@
...
+            # Recreate optimizer AFTER DDP wrapping
+            model._optimizer = None  # Clear cached optimizer
+            model.optimizer = torch.optim.Adam(model.model.parameters(), lr=0.001)
 
     result: ExperimentResult = model.train(dataloader, config.epochs)
 
     metrics = {
         "throughput_mibs": result["volume"] / result["training_duration_s"],
         "training_duration_s": result["training_duration_s"],
+        "volume_mibs": result["volume"],
```
jet-tong: Is this volume or volume mibs? Perhaps volume is not the best name for this - would total_samples reflect it better (or other alts?) (Perhaps "volume" could also be updated to e.g. "samples" or "num_samples" or etc?)

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -381,25 +391,27 @@ def create_fsspec_dataset(
...
         sampler=sampler,
         shuffle=False,
-        drop_last=True,
+        drop_last=use_drop_last,
         num_workers=num_workers,
         collate_fn=default_collate,
-        pin_memory=False,
-        persistent_workers=True,
+        pin_memory=torch.cuda.is_available(),
+        persistent_workers=num_workers > 0,
```
jet-tong: I understand `drop_last` is required, but perhaps `pin_memory` and `persistent_workers` are not required for this PR, which we could drop instead of making them dependent on cuda / num_workers silently? We could also expose these to the config in yaml too and set it to the PyTorch defaults for user-friendliness, if we want to keep them.

s3torchbenchmarking/src/s3torchbenchmarking/models.py
```diff
@@ -149,9 +149,13 @@ def capped_loader(self, loader):
             yield from loader
             return
         world = dist.get_world_size()
+        rank = dist.get_rank()
+
+        logger.info(
```
jet-tong: I think all workers use this loader and would print info for all ranks - do you know what the logs look like during the benchmarks so we know if logging them for all workers would be too verbose or not? Should also make sure there's no race conditions on the logs that make the outputs uglier. (Idea - perhaps place in debug)

(Same with logger.info below)

s3torchbenchmarking/src/s3torchbenchmarking/models.py
```diff
@@ -196,6 +208,8 @@ def train(self, dataloader: DataLoader, epochs: int) -> ExperimentResult:
             begin_training = perf_counter()
             for epoch in range(epochs):
                 begin_epoch = time.perf_counter()
+                if hasattr(dataloader.sampler, "set_epoch"):
```
jet-tong: I noticed this is new - may I ask why this is needed? 

(A comment would be good here if it's non-obvious, if it's trivial then no need)

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -381,25 +391,27 @@ def create_fsspec_dataset(
...
 
-def make_dataloader(dataset: Dataset, num_workers: int, batch_size: int, sampler=None):
+def make_dataloader(
+    dataset: Dataset, num_workers: int, batch_size: int, sampler=None, drop_last="auto"
+):
+    use_ddp = dist.is_initialized() and dist.get_world_size() > 1
+    if drop_last == "auto":
+        use_drop_last = use_ddp
+    else:
+        use_drop_last = str(drop_last).lower() == "true"
```
jet-tong, line 394: So 'False' input will convert to true?! We need to revise this logic or revise the input in the yaml file :)

s3torchbenchmarking/conf/dataset.yaml
```diff
@@ -40,4 +40,17 @@ hydra:
...
+      #   - S3MapDataset uses DistributedSampler
+      #   - capped_loader ensures all ranks process the same number of batches
+      # When num_gpus = 1 (or model=entitlement):
+      #   - Single process, no DDP overhead
+      +num_gpus: 1
+      # Whether to drop the last incomplete batch from each worker.
+      # "auto" (default): True for multi-GPU (required for DDP synchronization),
+      #                    False for single-GPU (all data is processed).
+      # Can be overridden to "true" or "false" explicitly.
+      +drop_last: auto
```
jet-tong: A few comments here:
- Perhaps we need to add something to the benchmark README to explain this behaviour, something simple to describe the effect setting num_gpus > 1 would have that is very non-obvious (routes to DDP and forces some config)
- num_gpus
  - Perhaps put num_gpus = 1 first to emphasize that's the default
  - Maybe put capped_loader comment first (iirc capped_loader splits samples of S3IterableDataset / S3MapDataset with different strategies), and iterable/map as sub-bullet points?
- drop_last:
  - Perhaps multi-gpu requires drop_last true needs to be emphasized
  - I don't think false works based on logic below?

From what I understand we want to make this clear (mermaid with some AI help so edits may be needed), perhaps we should make the comments clearer to convey this or add some stuff to the benchmark README:
```mermaid
flowchart TD
    A[run_experiment] --> B{model = entitlement?}
    B -->|Yes, I/O only| C[Force num_gpus=1]
    B -->|No| D{num_gpus > 1?}
    C --> E[Single Rank benchmark]
    D -->|No| E
    D -->|Yes| F[DDP: one process per GPU, requires drop_last=true so some data might be skipped; see metrics output for actual number of samples used]

    F --> H{Dataset type?}
    H -->|Map-style| I["DistributedSampler shards data
    capped_loader caps to min(len) across ranks"]
    H -->|Iterable| J["enable_sharding splits data
    capped_loader stops all when any rank exhausts"]
```

---
Mermaid source to make copying easier.
```
flowchart TD
    A[run_experiment] --> B{model = entitlement?}
    B -->|Yes, I/O only| C[Force num_gpus=1]
    B -->|No| D{num_gpus > 1?}
    C --> E[Single Rank benchmark]
    D -->|No| E
    D -->|Yes| F[DDP: one process per GPU, requires drop_last=true so some data might be skipped; see metrics output for actual number of samples used]

    F --> H{Dataset type?}
    H -->|Map-style| I["DistributedSampler shards data
    capped_loader caps to min(len) across ranks"]
    H -->|Iterable| J["enable_sharding splits data
    capped_loader stops all when any rank exhausts"]
```

jet-tong (commented):
Thanks @tadiwa-aizen for taking this - left a few comments on the config comments + usage to address and a bunch of small nits on style. 

@MSH4R1F Thanks for all your work on this benchmark feature! Sorry for leaving this open for so long, will be working with @tadiwa-aizen to merge this in.

CHANGELOG.md
```diff
@@ -11,13 +11,13 @@
...
 
 ### Other changes
-* Extended benchmarks to support multi-GPU Dataloading
 * Add seekable() method in S3Reader to eliminate tensor copies during DCP loading (#359)
 * Add load ordering optimization to S3StorageReader for sequential access patterns (#372)
 * Add os, arch, and PyTorch version to user agent string (#397)
 * Add benchmark to run DCP Loading Workloads (#357)
 * Add thread_count parameter to S3StorageWriter (#370)
 * Add macOS x86_64 and Python 3.8 deprecation warnings (#400)
+* Extended benchmarks to support multi-GPU Dataloading (#356)
```
jet-tong: Will likely move to new TBD column for this line now since we've released v1.5.0.

I haven't looked through all the changes yet, but it would be helpful if you resolve / comment on previous comments to note they're addressed. I had a quick look but it seems some comment locations weren't changed, e.g. `if drop_last == "auto"` line (not sure if addressed elsewhere though, pinged that comment also from this review).

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
jet-tong: Quick ping on this - is this comment addressed?

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -6,48 +6,130 @@
...
+    finally:
+        dist.destroy_process_group()
 
 
 # TODO: add Structured Config (https://hydra.cc/docs/tutorials/structured_config/intro/)
 @hydra.main(version_base=None)
 def run_experiment(config: DictConfig) -> dict:
+
+    num_gpus = torch.cuda.device_count()
+    if num_gpus > 1 and not dist.is_initialized():
```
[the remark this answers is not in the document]
tadiwa-aizen: Resolved

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
tadiwa-aizen: To asnwer the original comment .. no — this just checks if the value equals "true", so false stays false; it never flips it to true.

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
tadiwa-aizen: Done — removed the intermediate variable; drop_last is now fetched inline in the make_dataloader call like num_workers/batch_size.

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -72,36 +70,30 @@ def run_experiment(config: DictConfig) -> dict:
         os.environ["CUDA_VISIBLE_DEVICES"] = ",".join(map(str, range(num_gpus)))
         torch.cuda.empty_cache()
 
-    # In the case of multiple GPU training we run run_ddp_process using mp.spawn for each of the ranks, which calls run_benchmark_experiment separately
-    # If single-rank training then it defaults to run_benchmark_experiment
```
jet-tong: np: could keep this comment (with updated function name) since it explains the split fairly well.

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -408,10 +397,6 @@ def make_dataloader(
         drop_last=use_drop_last,
         num_workers=num_workers,
         collate_fn=default_collate,
-        pin_memory=torch.cuda.is_available(),
-        persistent_workers=num_workers > 0,
-        prefetch_factor=2 if num_workers > 0 else None,
```
jet-tong: Good find on these improvements, noted we dropped it to not interfere with testing.

qq: did we find out how much better these are to our benchmarks? If yes it might be worth adding pin memory / persistent workers / prefetch factor configs into the dataloading benchmarks. (Just added comment to not forget about these)

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -225,12 +380,21 @@ def create_fsspec_dataset(
     return dataset.map(load_sample)
 
 
-def make_dataloader(dataset: Dataset, num_workers: int, batch_size: int):
+def make_dataloader(
+    dataset: Dataset, num_workers: int, batch_size: int, sampler=None, drop_last="auto"
+):
+    use_ddp = dist.is_initialized() and dist.get_world_size() > 1
+    if drop_last == "auto":
```
jet-tong: np: can use `str(drop_last).lower()` like below in case of 'Auto' for example.

s3torchbenchmarking/README.md
```diff
@@ -93,6 +93,27 @@ Here are some sample dataset configurations that we ran our benchmarks against:
 s3torch-datagen -n 100k --shard-size 128MiB --s3-bucket my-bucket --region us-east-1
 ```
 
+## Multi-GPU dataset benchmarks
```
jet-tong: Thanks for adding the docs!

This placement seems a bit awkward - perhaps we can place it at the bottom since it's additional configuration notes that doesn't need to be read before vim ./conf/dataset.yaml  step (and I don't think it needs to be read in 1 GPU / entitlement usecases either which is more representative of potential throughput). (Or as a ### section after ## Running the benchmarks if we really need to place it in the middle)

s3torchbenchmarking/README.md
```diff
@@ -93,6 +93,27 @@ Here are some sample dataset configurations that we ran our benchmarks against:
...
+
+Setting `num_gpus > 1` in `conf/dataset.yaml` enables multi-GPU benchmarking via DDP (DistributedDataParallel).
```
jet-tong: nit: Can link to [DistributedDataParallel](https://docs.pytorch.org/tutorials/intermediate/ddp_tutorial.html).

s3torchbenchmarking/src/s3torchbenchmarking/models.py
```diff
@@ -153,11 +206,19 @@ def train(self, dataloader: DataLoader, epochs: int) -> ExperimentResult:
             begin_training = perf_counter()
             for epoch in range(epochs):
                 begin_epoch = time.perf_counter()
-                logger.info("Epoch #%i/%i", epoch, epochs - 1)
-                for batch_idx, (data, target) in enumerate(dataloader):
-                    logger.debug("Batch #%i", batch_idx)
+                if hasattr(dataloader.sampler, "set_epoch"):
+                    # DistributedSampler needs epoch to reseed its shuffle each epoch
```
jet-tong: Thanks for adding the comment but am still slightly confused, saw shuffle=False above. 

```python
    return DataLoader(
        dataset=dataset,
        batch_size=batch_size,
        sampler=sampler,
        shuffle=False,
```

s3torchbenchmarking/conf/dataset.yaml
```diff
@@ -39,4 +39,13 @@ hydra:
...
       dataloader.s3reader.type: sequential
       # buffer_size (bytes): only used with range_based s3reader
-      dataloader.s3reader.buffer_size: 8*1024*1024
\ No newline at end of file
+      dataloader.s3reader.buffer_size: 8*1024*1024
+      # Number of GPUs to use (default: 1 = single process, no DDP).
+      # Setting num_gpus > 1 enables DDP via mp.spawn. See README for details.
+      # Note: model=entitlement always forces num_gpus=1 (I/O-only, no training).
+      +num_gpus: 1
+      # Whether to drop the last incomplete batch.
```
jet-tong: Slightly confusing on what last incomplete batch comes from at first read. 

Suggestion:
```yaml
# Whether each rank drops its final batch when it has fewer than batch_size samples.
...
```

s3torchbenchmarking/conf/dataset.yaml
```diff
@@ -39,4 +39,13 @@ hydra:
...
+      # "auto" (default): True when num_gpus > 1, False otherwise.
+      # Multi-GPU requires drop_last=true for DDP synchronization; setting it to
+      # false with num_gpus > 1 will cause training to hang.
```
jet-tong: qq: does capped_loader mean it wouldn't hang anyways, or would it still hang with this combination?

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -23,17 +30,118 @@
...
 def run_experiment(config: DictConfig) -> dict:
+
+    num_gpus = (
+        config.num_gpus if hasattr(config, "num_gpus") else torch.cuda.device_count()
+    )
+    if not isinstance(num_gpus, int) or num_gpus <= 0:
+        raise ValueError(
+            "Invalid number of GPUs detected. Please specify a number of GPUs to use."
+        )
+    elif num_gpus > torch.cuda.device_count():
```
jet-tong: (AI Flagged): Noticed when using CPU-only hosts (like c / m instances not g instances) it would fail here. 

Perhaps add `elif num_gpus > 1 and num_gpus > torch.cuda.device_count():` to fix. 

And perhaps above mention of:`config.num_gpus if hasattr(config, "num_gpus") else torch.cuda.device_count()` -> `config.num_gpus if hasattr(config, "num_gpus") else 1`.

tadiwa-aizen (dismissed):

s3torchbenchmarking/conf/dataset.yaml
```diff
@@ -44,8 +44,11 @@ hydra:
...
       # Note: model=entitlement always forces num_gpus=1 (I/O-only, no training).
       +num_gpus: 1
-      # Whether to drop the last incomplete batch.
+      # Whether each rank drops its final batch when it has fewer than batch_size samples.
       # "auto" (default): True when num_gpus > 1, False otherwise.
-      # Multi-GPU requires drop_last=true for DDP synchronization; setting it to
-      # false with num_gpus > 1 will cause training to hang.
+      # This won't hang if set to false under multi-GPU: capped_loader already
+      # equalises the step count across ranks (map: min(len); iterable: stop when
+      # any rank exhausts), and that step-count sync is what prevents DDP hangs.
```
jet-tong: This addresses my comment on whether it would hang, but perhaps too verbose to keep in user-facing comment?

jet-tong: Perhaps we can also refer to benchmark README on this one too.

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
```diff
@@ -42,15 +153,32 @@ def run_experiment(config: DictConfig) -> dict:
...
+
+        if model.model is not None:
+            model.model = model.model.to(device_id)
+            model.device = torch.device(f"cuda:{device_id}")
+            model.model = torch.nn.parallel.DistributedDataParallel(
+                model.model, device_ids=[rank], output_device=rank
+            )
+            # Recreate optimizer AFTER DDP wrapping
+            model._optimizer = None  # Clear cached optimizer
+            model.optimizer = torch.optim.Adam(model.model.parameters(), lr=0.001)
```
jet-tong: Noted a bit late that this would mean hard coding the learning rate, not sure how best to fix at first glance...

This problem only hits when `dist.is_initialized` and learning rate is low priority when doing dataloading benchmarks. (looking back we might have wanted to explicitly enter this path if we entered `run_distributed_experiment` instead of `run_benchmark_experiment` instead of detecting dist.is_initialized)

s3torchbenchmarking/src/s3torchbenchmarking/models.py
```diff
@@ -20,6 +20,7 @@
 from torch.nn import Module
 from torch.utils.data.dataloader import DataLoader
 from torchvision.transforms import v2  # type: ignore
+import itertools
```
jet-tong: np (noticed late): this should move to top as 1st import (alphabetically first standard lib)

s3torchbenchmarking/src/s3torchbenchmarking/dataset/benchmark.py
tadiwa-aizen: Noted, dropped the hardcoded lr — it's already set in Model.optimizer, so I just clear the cache and re-access it to rebuild over the wrapped params. Also removed _optimizer = None since it wasn't doing anything.

jet-tong: Nice!

jet-tong (approved):

PR #357, opened 2025-08-22
Add Load Benchmarking Support for DCP DDP and FSDP

MSH4R1F:
# Add Load Benchmarking Support for DCP DDP and FSDP

## Summary
This PR extends the distributed checkpointing (DCP) benchmarking suite to support both save and load operations for DDP and FSDP training scenarios. Previously, the benchmarks only supported checkpoint saving; now users can benchmark checkpoint loading performance as well.

### 🚀 **New Features**
- **Load Benchmarking**: Added `load_benchmark.py` files for both DCP DDP and FSDP scenarios
- **Unified Script Interface**: Extended `run_benchmarks.sh` with `-l` flag to switch between save (default) and load operations
- **Configuration Support**: Created dedicated load configuration files (`dcp_ddp_load.yaml`, `dcp_fsdp_load.yaml`) with `checkpoint.suffix` parameter

### 📁 **Files Added/Modified**

**New Files:**
- `src/s3torchbenchmarking/dcp_ddp/load_benchmark.py` - DDP load benchmarking implementation
- `src/s3torchbenchmarking/dcp_fsdp/load_benchmark.py` - FSDP load benchmarking implementation
- `conf/dcp_ddp_load.yaml` - Load configuration for DDP benchmarks
- `conf/dcp_fsdp_load.yaml` - Load configuration for FSDP benchmarks

**Modified Files:**
- `utils/run_benchmarks.sh` - Added `-l` flag support and config file selection logic
- `src/s3torchbenchmarking/dcp_common.py` - Enhanced with `get_reader()` function for load operations
- `src/s3torchbenchmarking/dcp_ddp/README.md` - Updated documentation
- `src/s3torchbenchmarking/dcp_fsdp/README.md` - Updated documentation

### 🔧 **Technical Implementation**
- **Reader Support**: Added `get_reader()` function in `dcp_common.py` to instantiate S3 and filesystem readers
- **Config-Driven Suffix**: Load benchmarks get checkpoint suffix from config (`cfg.checkpoint.suffix`) instead of parameters
- **Backward Compatibility**: Default behavior unchanged - save operations work exactly as before

### 📊 **Usage**
```bash
# Save benchmarks (default behavior)
./utils/run_dcp_ddp_benchmarks.sh
./utils/run_dcp_fsdp_benchmarks.sh

# Load benchmarks (new functionality)
./utils/run_dcp_ddp_benchmarks.sh -l
./utils/run_dcp_fsdp_benchmarks.sh -l

### DIFF
--- a/s3torchbenchmarking/conf/dcp_ddp_load.yaml
+++ b/s3torchbenchmarking/conf/dcp_ddp_load.yaml
@@ -1,0 +1,31 @@
+defaults:
+  - hydra/callbacks/collate_results
+  - aws/dynamodb # save run results to DynamoDB -- comment me if not required
+  - _self_
+
+# S3 bucket to use to save checkpoints.
+# NOTE: a non-existing bucket will fail the benchmarks.
+s3:
+  region: # e.g., eu-west-1
+  uri: # e.g., s3://my-bucket/
+# Number of iterations for "saving" a model's checkpoint.
+# NOTE: this does not affect model training, as no actual training occurs in these benchmarks.
+epochs: 4
+
+hydra:
+  mode: MULTIRUN
+  sweep:
+    dir: multirun/${hydra.job.config_name}/${now:%Y-%m-%d_%H-%M-%S}
+  sweeper:
+    params:
+      # Short name of a pre-trained model (from Hugging Face), listed in `models.py`.
+      +model: vit-base, T0_3B
+      # Type of Torch distributed backend (valid options: "nccl", "gloo").
+      +backend: nccl
+      # Number of workers.
+      +world_size: 4
+      # Number of threads to use for saving the checkpoints.
+      +thread_count: 4
+      # Checkpoint storage location (valid options: "disk", "s3").
+      +checkpoint.storage: disk, s3
+      +checkpoint.suffix: 
--- a/s3torchbenchmarking/conf/dcp_fsdp_load.yaml
+++ b/s3torchbenchmarking/conf/dcp_fsdp_load.yaml
@@ -1,0 +1,39 @@
+defaults:
+  - hydra/callbacks/collate_results
+  - aws/dynamodb # save run results to DynamoDB -- comment me if not required
+  - _self_
+
+# S3 bucket to use to save checkpoints.
+# NOTE: a non-existing bucket will fail the benchmarks.
+s3:
+  region:  # e.g., eu-west-1
+  uri:   # e.g., s3://my-bucket/
+# Number of iterations for "saving" a model's checkpoint.
+# NOTE: this does not affect model training, as no actual training occurs in these benchmarks.
+epochs: 4
+
+hydra:
+  mode: MULTIRUN
+  sweep:
+    dir: multirun/${hydra.job.config_name}/${now:%Y-%m-%d_%H-%M-%S}
+  sweeper:
+    params:
+      # Short name of a pre-trained llama v2 model (valid options: "L7b", "L13b", "L30b", "L65b", "L70b").
+      +model: L7b, L13b, L30b
+      # Type of Torch distributed backend (valid options: "nccl", "gloo").
+      +backend: nccl
+      # Number of workers.
+      +world_size: 4
+      # Number of threads to use for saving the checkpoints.
+      +thread_count: 4
+      # Checkpoint storage location (valid options: "disk", "s3").
+      +checkpoint.storage: disk, s3
+      # Sharding strategy (valid options: "full", "hybrid").
+      +checkpoint.sharding_strategy: full
+      # Controls whether files are forcibly synced to disk (only relevant for "disk" storage).
+      # NOTE: We disabled this option to improve performance since FSDP checkpointing with
+      # forced syncing (maximum durability) was significantly slower than storage throughput.
+      # This setting has no effect when using "s3" storage.
+      +checkpoint.sync_files: false
+      +checkpoint.suffix: 
+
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: s3torchbenchmarking/conf/dcp_ddp_load.yaml [hunk 1]
```diff
@@ -0,0 +1,31 @@
+defaults:
```
jet-tong, line 1: Is there a chance we could use same config for save/load? So we don't need to fill in region/uri for every config file.

I did notice most of the config is overlapped, but suffix needs changes (and we would sometimes use world size 1 for loading benchmarks)

s3torchbenchmarking/conf/dcp_ddp_load.yaml
```diff
@@ -0,0 +1,31 @@
...
+      +model: vit-base, T0_3B
+      # Type of Torch distributed backend (valid options: "nccl", "gloo").
+      +backend: nccl
+      # Number of workers.
+      +world_size: 4
+      # Number of threads to use for saving the checkpoints.
+      +thread_count: 4
+      # Checkpoint storage location (valid options: "disk", "s3").
+      +checkpoint.storage: disk, s3
+      +checkpoint.suffix:
```
jet-tong, line 31: We can do it like dcp_fsdp.yaml to prompt users to update it, with an example of the random suffix generated in the save script:
```
s3:
  region: ??? # e.g., eu-west-1
  uri: ???    # e.g., s3://my-bucket/
```

s3torchbenchmarking/conf/dcp_fsdp_load.yaml
```diff
@@ -23,9 +23,9 @@ hydra:
       # Type of Torch distributed backend (valid options: "nccl", "gloo").
       +backend: nccl
       # Number of workers.
-      +world_size: 8
+      +world_size: 4
       # Number of threads to use for saving the checkpoints.
-      +thread_count: 8
+      +thread_count: 4
```
jet-tong, line 28: Why change from 8 to 4?

s3torchbenchmarking/conf/dcp_fsdp_load.yaml
```diff
@@ -35,4 +35,5 @@ hydra:
       # forced syncing (maximum durability) was significantly slower than storage throughput.
       # This setting has no effect when using "s3" storage.
       +checkpoint.sync_files: false
+      +checkpoint.suffix:
```
jet-tong, line 38: Same as above in dcp_ddp

s3torchbenchmarking/src/s3torchbenchmarking/dcp_common.py
```diff
@@ -49,6 +50,19 @@ def get_writer(cfg: DictConfig, suffix: str) -> FileSystemWriter:
     raise ValueError(f"Storage writer {cfg.checkpoint.storage} not supported")
 
 
+def get_reader(cfg: DictConfig, suffix: str) -> FileSystemReader:
```
jet-tong, line 53: Below we have:
```
    # Override random suffix with suffix from config
    suffix = cfg.checkpoint.suffix
    storage_reader = get_reader(cfg, suffix)
```

That would mean we don't have to pass suffix to get_reader if we already pass cfg, we could just use cfg.checkpoint.suffix directly.

jet-tong (commented):
Nice - having 'save' to differentiate between save and load benchmarks is great.

Backwards compatibility is important, but clear naming/structure is just as important.

Will have another look later. Good work!

s3torchbenchmarking/src/s3torchbenchmarking/dcp_fsdp/load_benchmark.py
```diff
@@ -0,0 +1,141 @@
...
+    return benchmark_common_runner(cfg, run_fsdp_load, (cfg,))
+
+
+def run_fsdp_load(
+    rank: int,
+    cfg: DictConfig,
+    suffix: str,
+    load_timestamps: Queue,
+):
+    """Execute the actual code for checkpoint saving.
```
jet-tong, line 49: DCP loading here

s3torchbenchmarking/src/s3torchbenchmarking/dcp_fsdp/load_benchmark.py
```diff
@@ -0,0 +1,141 @@
...
+@hydra.main(version_base=None)
+def run_benchmark(cfg: DictConfig) -> dict:
+    """DCP load benchmarks entry point."""
+    return benchmark_common_runner(cfg, run_fsdp_load, (cfg,))
+
+
+def run_fsdp_load(
+    rank: int,
+    cfg: DictConfig,
+    suffix: str,
```
jet-tong, line 46: same here - suffix probably in cfg already?

s3torchbenchmarking/conf/dcp_fsdp_load.yaml
```diff
@@ -0,0 +1,39 @@
...
+epochs: 4
+
+hydra:
+  mode: MULTIRUN
+  sweep:
+    dir: multirun/${hydra.job.config_name}/${now:%Y-%m-%d_%H-%M-%S}
+  sweeper:
+    params:
+      # Short name of a pre-trained llama v2 model (valid options: "L7b", "L13b", "L30b", "L65b", "L70b").
+      +model: L7b, L13b, L30b
```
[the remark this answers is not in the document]
MSH4R1F: Yepp, right, because we are loading one specific type of model there should not be multiple options for model, and should leave the user to input which model type they are loaidng

s3torchbenchmarking/src/s3torchbenchmarking/dcp_common.py
MSH4R1F: Thanks!

PR #360, opened 2025-08-28
fix(s3writer): override closed property and block writes after close

jet-tong:
## Description

S3Writer changes:
- Override closed property to reflect _closed flag instead of BufferedIOBase.__closed
- Update writable() to return False after S3Writer is closed
- When writing on closed streams, use python/cpython/Lib/_pyio.py _checkClosed() method to raise ValueError

These fixes improve correctness without changing normal behaviour.

## Additional context

Our [S3Writer](https://github.com/awslabs/s3-connector-for-pytorch/blob/main/s3torchconnector/src/s3torchconnector/s3writer.py) extends io.BufferedIOBase (in [_pyio.py](https://github.com/python/cpython/blob/main/Lib/_pyio.py)), which has a [closed](https://github.com/python/cpython/blob/055827528fa50c9a7707792f5fe264c4e20c07e9/Lib/_pyio.py#L462) property that reflects __closed, and we did not override it to represent our attribute self._closed. PyTorch methods might be using these properties/methods - and we improve correctness in this PR to avoid possible inconsistencies.

## Testing

Added unit tests for these 3 changes. 

--------

### DIFF
--- a/s3torchconnector/src/s3torchconnector/s3writer.py
+++ b/s3torchconnector/src/s3torchconnector/s3writer.py
@@ -51,7 +51,9 @@ class S3Writer(io.BufferedIOBase):
 
         Raises:
             S3Exception: An error occurred accessing S3.
+            ValueError: If the writer is closed.
         """
+        self._checkClosed()  # from python/cpython/Lib/_pyio.py
         if isinstance(data, memoryview):
             data = data.tobytes()
         self.stream.write(data)
@@ -69,6 +71,14 @@ class S3Writer(io.BufferedIOBase):
                 self._closed = True
                 self.stream.close()
 
+    @property
+    def closed(self) -> bool:
+        """
+        Returns:
+            bool: Return whether the object is closed.
+        """
+        return self._closed
+
     def flush(self):
         """No-op"""
         pass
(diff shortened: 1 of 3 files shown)


s3torchconnector/src/s3torchconnector/s3writer.py
```diff
@@ -51,7 +51,9 @@ def write(
 
         Raises:
             S3Exception: An error occurred accessing S3.
+            ValueError: If the writer is closed.
         """
+        self._checkClosed()  # from python/cpython/Lib/_pyio.py
```
jet-tong, line 56: [Link to _pyio.py](https://github.com/python/cpython/blob/c779f2324df06563b4ba3d70d0941e619ccaf5ff/Lib/_pyio.py#L470)

Should check that this change
- doesn't cause regressions
- won't be masking stream errors

No e2e tests written for these changes specifically - I haven't found any suitable end to end scenarios to add here.

IsaevIlya (approved):
Looks good to me!

s3torchconnector/src/s3torchconnector/s3writer.py
jet-tong: No regressions found in dcp fsdp checkpointing benchmarks.
And since it only checks for S3Writer closed state, stream errors (e.g. underlying stream being closed unexpectedly) wouldn't be masked by _checkClosed() function.

PR #368, opened 2025-09-16
Shadow Copies 

MSH4R1F:
## Description
Shadow Copies automatically duplicates checkpoints across different S3 prefixes during the save process. A metadata file records the number of copies. During load, we assign each worker a specific copy to load from using round-robin, increasing the S3 throttling limit. This is especially needed for workloads where each worker needs to load the entire model.

#### 1. S3StorageWriter
- Creates multiple copies in parallel during checkpointing
- Updates metadata files to track checkpoint copies and their checkpoints
#### 2. S3StorageReader
- Fetches metadata to discover number of copies
- Distributes workers across prefixes using round-robin
- Enables parallel loading from multiple S3 locations for improved throughput

We extend this in a backwards-compatible way with default `num_copies = 1` reverting to original S3 Connector for Pytorch method.

### How to use?
For saving:
```py
writer = S3StorageWriter(
    region=REGION,
    path="CHECKPOINT_URI",
    num_copies = 8 # Creates 8 copies
)
dcp.save(state_dict, storage_writer = writer)
```

For loading (no change, automatically detected):
```
reader = S3StorageReader(
    region=REGION,
    path="CHECKPOINT_URI"
)
dcp.load(state_dict = state_dict, storage_reader = reader)
```
Output structure:
```
CHECKPOINT_URI
├── copy-1/
│   └── ├── __0_0.distcp
│       ├── __1_0.distcp
│       └── ...
├── copy-2/
│   └── ├── __0_0.distcp
│       ├── __1_0.distcp
│       └── ...
└── copy-3/
    └── ├── __0_0.distcp
        ├── __0_1.distcp
        └── ...
```

## Testing 
I've added a benchmarking suite to the 's3torchbenchmarking/dcp_fsdp' module for testing. It contains scripts that run our connector using Distributed Checkpointing and Distributed Loading via Slurm, allowing scaling across multiple nodes.

Instructions to reproduce these tests are included in [Testing instructions](https://github.com/MSH4R1F/s3-connector-for-pytorch/blob/main/s3torchbenchmarking/src/s3torchbenchmarking/dcp_fsdp/README.md). Testing and benchmarking has shown us that is one of the ways to reduce throttling from S3 and allows us to scale to hundreds of node without failing workloads

## Risks
We are overwriting the write_data method when there's more than only copy, which would couple the PR to specific PyTorch versions which is Pytorch 2.6+ after they refactor their write_data code, which is why Python 3.8 (which is restricted to Pytorch <=2.5) is failing tests.

We're planning to put this in a feature branch only, but this might require limiting PyTorch versions in pyproject.toml. 


jet-tong (commented):
Really cool feature! As called out below, due to write_data function being specific to some pytorch versions (likely 2.6+ - to check), we're putting it in feature branch only, and might want to add limitations to pytorch version in pyproject.toml.

Would appreciate some unit tests e.g. to make sure num_copies is passed correctly or edge cases, and other e2e tests e.g. for loading checkpoints (maybe with only 1 worker, to see if they only fetch assigned copy, and assert which copy it is - might be a tough write). 

For the Slurm scripts - awesome way to benchmark for large scale training workloads. Since this PR might only go to feature branch, probably worth a PR to main for the slurm scripts and adapt them to [save/load benchmarks in your other PR](https://github.com/awslabs/s3-connector-for-pytorch/pull/357) (excluding the shadow copy specific lines).

Good work!

jet-tong (approved):
Noted this feature doesn't support PyTorch < 2.6 (and hence Python 3.8 which uses PyTorch 2.4.1) since it overrides PyTorch `FileSystemWriter.write_data` method, which got updated in PyTorch 2.6. 

LGTM for merging to a feature branch. 

Good work! 🚀

MSH4R1F:
Thanks Jensen!

PR #370, opened 2025-09-22
Adding thread_count parameter to S3StorageWriter

MSH4R1F:
## Summary
Added optional thread_count parameter to S3StorageWriter to allow users to configure the number of I/O threads for write operations.

## Changes
• **S3StorageWriter**: Added thread_count parameter (defaults to 1) to constructor
• **Documentation**: Updated README example to show the new optional parameter

## Usage
```python
# Use default single thread
s3_storage_writer = S3StorageWriter(region=REGION, path=CHECKPOINT_URI)

# Configure multiple I/O threads for better performance
s3_storage_writer = S3StorageWriter(region=REGION, path=CHECKPOINT_URI, thread_count=4)
```

This enhancement allows users to optimize write performance by configuring the number of concurrent I/O threads based on their specific use case and infrastructure.

### DIFF
--- a/s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
+++ b/s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
@@ -271,6 +271,7 @@ class S3StorageWriter(FileSystemWriter):
         path: str,
         s3client_config: Optional[S3ClientConfig] = None,
         prefix_strategy: Optional[S3PrefixStrategyBase] = None,
+        thread_count: int = 1,
         **kwargs,
     ) -> None:
         """
@@ -287,6 +288,7 @@ class S3StorageWriter(FileSystemWriter):
         super().__init__(
             path=path,
             sync_files=False,  # FIXME: setting this to True makes the run to fail (L#333: `os.fsync(stream.fileno())`)
+            thread_count=thread_count,
             **kwargs,
         )
         self.fs = S3FileSystem(region, s3client_config=s3client_config)  # type: ignore
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 * Add seekable() method in S3Reader to eliminate tensor copies during DCP loading (#359)
 * Override S3Writer closed property and block writes after close (#360)
 
+### Other changes
+* Added thread_count parameter to S3StorageWriter 
+
 ## v1.4.3 (July 25, 2025)
 
 ### New features
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ REGION = "us-east-1"
 model = torchvision.models.resnet18()
 
 # Save distributed checkpoint to S3
-s3_storage_writer = S3StorageWriter(region=REGION, path=CHECKPOINT_URI)
+s3_storage_writer = S3StorageWriter(region=REGION, path=CHECKPOINT_URI) # optional thread_count parameter for number of I/O threads to write
 DCP.save(
     state_dict=model.state_dict(),
     storage_writer=s3_storage_writer,


s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
```diff
@@ -271,6 +271,7 @@ def __init__(
         path: str,
         s3client_config: Optional[S3ClientConfig] = None,
         prefix_strategy: Optional[S3PrefixStrategyBase] = None,
+        thread_count: int = 1,
         **kwargs,
     ) -> None:
         """
```
jet-tong, line 277: Probs need edit on docstring too, referencing [PyTorch docstring](https://github.com/pytorch/pytorch/blob/96a3afb8ec1b0d387a032ef58f75a1975e567d67/torch/distributed/checkpoint/filesystem.py#L605C13-L606C1):
```
thread_count (int): Number of IO threads to use to write. PyTorch filesystem defaults to 1. 
```
(or similar)

jet-tong (commented):
Provided some suggested changes.

We already test for thread_count in test_e2e_s3_file_system.py (could check if that's sufficient), but probably worth adding unit tests in test_s3_file_system.py too. I'm not sure if we need to add a new file for test_s3_storage_writer.py, but if we do:
```
#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
#  // SPDX-License-Identifier: BSD

import pytest
from s3torchconnector.dcp.s3_file_system import S3StorageWriter

TEST_REGION = "eu-east-1"
TEST_PATH = "s3://test-bucket/test-key"


@pytest.mark.parametrize("thread_count", [1, 4, 8, 16])
def test_s3storage_writer_thread_count_valid(thread_count):
    writer = S3StorageWriter(region=TEST_REGION, path=TEST_PATH, thread_count=thread_count)
    assert writer.thread_count == thread_count
```
which utilizes [this pytorch attribute](https://github.com/pytorch/pytorch/blob/96a3afb8ec1b0d387a032ef58f75a1975e567d67/torch/distributed/checkpoint/filesystem.py#L617) to test if it passes through correctly.

s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
MSH4R1F: Yep, apologies missed this

s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
```diff
@@ -271,6 +271,7 @@ def __init__(
         path: str,
         s3client_config: Optional[S3ClientConfig] = None,
         prefix_strategy: Optional[S3PrefixStrategyBase] = None,
+        thread_count: int = 1,
```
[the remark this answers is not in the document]
jet-tong: Hmm I can see pytorch only accepts int (not None) for `thread_count`. An alternative I have would be to build the super init arguments (instead of if else):

```
    super_kwargs = {
        "path": path,
        "sync_files": False,
        **kwargs
    }
    if thread_count is not None:
        super_kwargs["thread_count"] = thread_count
    
    super().__init__(**super_kwargs)
```

But I prefer your approach because:
- It's simpler
- `thread_count` is not expected to change default from 1
- It exposes the default `thread_count` to users clearly

s3torchconnector/tst/unit/dcp/test_s3_storage_writer.py
```diff
@@ -0,0 +1,23 @@
...
+
+@pytest.mark.parametrize("thread_count", [1, 2, 4, 8, 16])
+def test_s3storage_writer_thread_count(thread_count):
+    storage_writer = S3StorageWriter(
+        region=TEST_REGION, path=TEST_PATH, thread_count=thread_count
+    )
+    assert storage_writer.thread_count == thread_count
+
+
+def test_s3storage_writer_thread_count_none_defaults_to_one():
```
jet-tong, line 21: Maybe this should now be:
```suggestion
def test_s3storage_writer_thread_count_defaults_to_one():
```

s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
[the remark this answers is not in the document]
MSH4R1F: That's pretty smart. But yep, I don't think thread_count default will change so will resolve for now, thanks!

s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
```diff
@@ -283,6 +283,7 @@ def __init__(
             s3client_config (Optional[S3ClientConfig]): Optional S3ClientConfig with parameters for S3 client.
             prefix_strategy (Optional[S3PrefixStrategyBase]): Optional strategy for generating S3 prefixes to
                 optimize checkpoint organization and prevent throttling.
+            thread_count (int): Number of IO threads to use to write. We default to 1 emulating Pytorch defaults
```
jet-tong, line 286: nit: maybe just write `Defaults to 1`, and optionally add (PyTorch default) / (Parent class default)
```suggestion
            thread_count (int): Number of IO threads to use to write. Defaults to 1 (PyTorch default).
```

jet-tong (commented):
Also found out that this python file was lacking Class docstrings, so our [dcp sphinx docs](https://awslabs.github.io/s3-connector-for-pytorch/autoapi/s3torchconnector/dcp/index.html) are only showing parent class docstrings. Can be done in another PR if needed - we can check if other classes have the same problem too.

MSH4R1F:
> Also found out that this python file was lacking Class docstrings, so our [dcp sphinx docs](https://awslabs.github.io/s3-connector-for-pytorch/autoapi/s3torchconnector/dcp/index.html) are only showing parent class docstrings. Can be done in another PR if needed - we can check if other classes have the same problem too.

Would be good if we can identify all of these classes and do them all in one PR, I can create a task for that?

jet-tong (approved):
LGTM!

PR #372, opened 2025-09-29
perf(dcp): load ordering - sort load items by storage offset

jet-tong:
## Description

Implements load ordering optimization in S3StorageReader to improve PyTorch DCP loading performance by sorting checkpoint items by storage offset. This enables sequential access patterns, reducing load times by up to 26% when combined with ['seekable' S3Reader optimization](https://github.com/awslabs/s3-connector-for-pytorch/pull/359).

Note this does not prevent all backwards seeks - torch.load() will still make backwards seeks when reading each tensor object. 

Also contains a small `pyproject.toml` fix for cibuildwheels / Build Wheels workflow to test for all DCP tests, instead of only `test_e2e_s3_file_system.py`. 

Cherry-picked load ordering code from experimental PR https://github.com/awslabs/s3-connector-for-pytorch/pull/352, added additional unit tests, and updated docs/docstrings.

## Additional context

Problem: PyTorch DCP loads checkpoint items in arbitrary order, causing inefficient I/O patterns. For example, loading requests might access offsets like: 70KB → 350MB → 1.8GB → 80KB, creating large jumps throughout the file. Our sequential S3Reader buffers from offset 0 to the current position, so accessing offset 1.8GB requires downloading and buffering the entire 1.8GB, even if only small portions are needed.

Load Ordering Solution: sequentially loads items in local plan based on their actual offset in checkpoint shards, ensuring sequential access patterns and improves I/O efficiency. This also effectively addresses a PyTorch TODO comment about sorting requests by offset (in torch/distributed/checkpoint/filesystem.py read_data).

Benchmarks with loading a Llama 7B checkpoint (3.2GB * 8 shards) show that when combined with ['seekable' S3Reader optimization](https://github.com/awslabs/s3-connector-for-pytorch/pull/359), load times can be reduced by up to 26%. 

No breaking changes - the optimization is applied automatically during `dcp.load()` with no user configuration required.

## Related items

- Original experimental Save/Load ordering PR: https://github.com/awslabs/s3-connector-for-pytorch/pull/352
- Works with another DCP load optimization: https://github.com/awslabs/s3-connector-for-pytorch/pull/359

## Testing

- Unit / e2e tests in PR
- Integration tests automatically applies load ordering
- Validated performance with DCP load use cases

#### About the e2e test

- According to torch/distributed/checkpoint/storage.py, dcp.load() calls _load_state_dict (in state_dict_loader.py)
- This will call: read_metadata() > set_up_storage_reader() > prepare_local_plan() > prepare_global_plan() > read_data()
- As described in docstring, pytorch/torch/serialization.py load() function will call _is_zipfile(), which includes this read() call: f.read(len(local_header_magic_number)). This is followed by readinto() calls on the actual tensor. So tensor read follows pattern, e.g.:
```bash
  type=read, position=72500, size=4
  type=readinto, position=72500, size=8
  type=readinto, position=72500, size=3625
  type=readinto, position=76103, size=22
  ...
  type=read, position=76125, size=4
  type=readinto, position=76125, size=8
```
- Hence we can track read() call positions to determine if load ordering is being applied correctly.

For the short example, the read calls before and after Load Ordering:
- Before: [1641, 0, 6395, 3218, 9997, 8036]
- After: [0, 1641, 3218, 6395, 8036, 9997]

Note we could also use torchvision.models.resnet18(pretrained=False) as a model, but did not use due to torchvision package issues (and rely on external packages less for integration tests).

--------

### DIFF
--- a/s3torchconnector/tst/unit/dcp/test_s3_storage_reader.py
+++ b/s3torchconnector/tst/unit/dcp/test_s3_storage_reader.py
@@ -1,0 +1,120 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  // SPDX-License-Identifier: BSD
+
+from typing import Dict, Any
+from unittest.mock import Mock
+from hypothesis import given, assume
+from hypothesis.strategies import composite, integers, lists
+
+import torch
+import torch.distributed.checkpoint as dcp
+from torch.distributed.checkpoint.planner import LoadPlan, ReadItem, LoadItemType
+from torch.distributed.checkpoint.metadata import (
+    Metadata,
+    MetadataIndex,
+    TensorStorageMetadata,
+    ChunkStorageMetadata,
+)
+
+from s3torchconnector.dcp import S3StorageReader
+
+TEST_REGION = "eu-east-1"
+TEST_PATH = "s3://test-bucket/test-checkpoint/"
+
+
+@composite
+def load_plan_with_offsets(draw):
+    """Generate LoadPlan with random offsets."""
+    offsets = draw(lists(integers(0, 10_000_000), min_size=0, max_size=10_000))
+
+    storage_data = {}
+    items = []
+
+    for i, offset in enumerate(offsets):
+        metadata_index = MetadataIndex(fqn=f"item{i}", offset=torch.Size([0]), index=0)
+
+        # Mock storage info
+        storage_data[metadata_index] = Mock(
+            offset=offset,
+            length=draw(integers(1000, 50000)), # DCP requires length - use random integers
+            relative_path=f"__{draw(integers(0, 7))}_0.distcp",
+        )
+
+        items.append(
+            Mock(spec=ReadItem, storage_index=metadata_index, type=LoadItemType.TENSOR)
+        )
+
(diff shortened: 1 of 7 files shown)


s3torchconnector/tst/unit/dcp/test_s3_storage_reader.py
```diff
@@ -0,0 +1,126 @@
...
+    # Create state_dict matching the metadata structure
+    state_dict = {f"item{i}": torch.zeros(10) for i in range(len(load_plan.items))}
+
+    # 1. In torch/distributed/checkpoint/state_dict_loader.py: dcp.load() calls _load_state_dict;
+    # 2. According to torch/distributed/checkpoint/storage.py StorageWriter docstring, _load_state_dict() calls:
+    #    read_metadata() > set_up_storage_reader() > prepare_local_plan() > prepare_global_plan() > read_data()
+    dcp.load(state_dict, storage_reader=s3_storage_reader)
+
+    # When read_data is called, verify prepare_local_plan was called and sorted the items
+    sorted_plan = s3_storage_reader.read_data.call_args[0][0]  # First arg is the plan
```
muddyfish, line 116: Perhaps an `assert isinstance` here would help verify this doesn't change

PR #374, opened 2025-10-03
ci: migrate from macos-13 to macos-15-intel image

jet-tong:
## Description

According to https://github.com/actions/runner-images/issues/13046
- macos-13 will have Github Actions brownouts starting November 4th, and will be unsupported by December 4th. 

This PR:
- upgrades from macos-13 to macos-15-intel.
- Note macos-15-intel planned to end August 2027.

## Additional context

- [ ] I have updated the CHANGELOG or README if appropriate

## Related items

https://github.com/actions/runner-images/issues/13046
> [macOS] The macOS 13 Ventura based runner images will begin deprecation on September 22nd and will be fully unsupported by December 4th for GitHub and ADO #13046

## Testing

Build Wheels workflow: https://github.com/awslabs/s3-connector-for-pytorch/actions/runs/18220092432

--------
By submitting this pull request, I confirm that my contribution is made under the terms of BSD 3-Clause License and I agree to the terms of the [LICENSE](https://github.com/awslabs/s3-connector-for-pytorch/blob/main/LICENSE).

PR #376, opened 2025-10-03
fix(test): fix AWS_IO_DNS_INVALID_NAME in multiprocess tests

jet-tong:
## Description

- Add `multiprocessing_context=mp.get_context()` to force DataLoader to use the same context as the main process (i.e. spawn instead of fork on Darwin) for multiprocessing e2e tests, 
- Prevents S3 client fork handlers from corrupting AWS CRT DNS resolver threads on macOS GitHub runners.

Alternatively, we can also consider changing behaviour on S3Client's fork handlers specifically on Darwin platform.

#### Error Context

Tests were failing with `AWS_IO_DNS_INVALID_NAME` errors on macos-14 (ARM64) GitHub Actions runners during our Build Wheels workflow, particularly in `test_distributed_training.py`. These errors were intermittent, across 5 macos-14 runners each running 40+ tests, typically 1-2 fails. We could resolve these failures by retrying the Build Wheels workflow 1-2 times for every run. 

The error re-appeared in PR [#374](https://github.com/awslabs/s3-connector-for-pytorch/pull/374). The error stack traces showed `popen_fork.py` being used, which was suspicious since we were explicitly setting `torch.multiprocessing.set_start_method('spawn')`, and disabled fork / forkserver methods for Darwin. 

#### Root Cause (Still Investigating)

PyTorch DataLoader source code (torch/utils/data/dataloader.py) showed that when `multiprocessing_context=None` (the default), DataLoader ignores the global `torch.multiprocessing.set_start_method()` setting and uses its own default context. This means DataLoader was using fork internally, triggering the S3 client's fork handlers.

When DataLoader forks workers, S3 client fork handlers (registered via `os.register_at_fork()` as a workaround for CRT multithreading issues in PR [#320](https://github.com/awslabs/s3-connector-for-pytorch/pull/320)) call `join_all_managed_threads()` to clean up AWS CRT background threads, but this corrupts active DNS resolver threads and causes `AWS_IO_DNS_INVALID_NAME` errors.

## Related items

- Related PR with errors: https://github.com/awslabs/s3-connector-for-pytorch/pull/374
- PR for workaround for fork: https://github.com/awslabs/s3-connector-for-pytorch/pull/320

## Testing

Build Wheels workflow: https://github.com/awslabs/s3-connector-for-pytorch/actions/runs/18225016406

--------

### DIFF
--- a/s3torchconnector/tst/e2e/test_distributed_training.py
+++ b/s3torchconnector/tst/e2e/test_distributed_training.py
@@ -79,7 +79,11 @@ def dataloader_for_map(
     )
     sampler = DistributedSampler(dataset)
     dataloader = DataLoader(
-        dataset, batch_size=batch_size, num_workers=num_workers, sampler=sampler
+        dataset,
+        batch_size=batch_size,
+        num_workers=num_workers,
+        sampler=sampler,
+        multiprocessing_context=mp.get_context(),
     )
     return dataloader
 
@@ -93,7 +97,12 @@ def dataloader_for_iterable(
         enable_sharding=True,
         reader_constructor=reader_constructor,
     )
-    dataloader = DataLoader(dataset, batch_size=batch_size, num_workers=num_workers)
+    dataloader = DataLoader(
+        dataset,
+        batch_size=batch_size,
+        num_workers=num_workers,
+        multiprocessing_context=mp.get_context(),
+    )
     return dataloader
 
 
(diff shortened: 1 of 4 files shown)


jet-tong:
`test_logging.py` `subprocess.Popen` could also cause the error. 

We can also consider changing behaviour on S3Client's fork handlers specifically on Darwin platform.

jet-tong:
Closing PR for now since this method doesn't seem to fix the root cause.

PR #378, opened 2025-10-07
feat(dcp): list of ranges reader for DCP partial loading

jet-tong:
## Description

ListOfRangesS3Reader for DCP Partial Loading: Implements ListOfRangesS3Reader to optimize PyTorch Distributed Checkpoint (DCP) partial loading by only fetching required byte ranges instead of entire objects.

- ListOfRangesS3Reader: Coalesces nearby ranges and manages multiple ranged streams per object
- Enhanced SequentialS3Reader: Added start_offset / end_offset support for ranged reads (This blurs boundary between sequential/range-based readers - to check if this is the best way to implement)
- DCP Integration: S3StorageReader automatically injects range metadata from DCP load plans when providing `dcp_list_of_ranges()` as reader_constructor

Usage:
```python
reader_constructor = S3ReaderConstructor.dcp_list_of_ranges()
storage_reader = S3StorageReader(region, path, reader_constructor=reader_constructor)
```

Optimized for partial DCP loading where only specific items are needed from large distributed checkpoint files.

## Additional context

## Testing

--------

### DIFF
--- a/s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
+++ b/s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
@@ -28,7 +28,11 @@ import torch
 
 from s3torchconnector._s3client import S3Client
 from s3torchconnector._s3dataset_common import parse_s3_uri
-from ..s3reader import S3ReaderConstructor, S3ReaderConstructorProtocol
+from ..s3reader import (
+    S3ReaderConstructor,
+    DCPListOfRangesConstructor,
+    S3ReaderConstructorProtocol,
+)
 from .. import S3ClientConfig
 from .s3_prefix_strategy import S3PrefixStrategyBase, DefaultPrefixStrategy
 from .._user_agent import UserAgent
@@ -79,7 +83,6 @@ class S3FileSystem(FileSystemBase):
         self,
         path: Union[str, os.PathLike],
         mode: str,
-        reader_constructor: Optional[S3ReaderConstructorProtocol] = None,
     ) -> Generator[io.IOBase, None, None]:
         """
         Create a stream for reading or writing to S3.
@@ -102,18 +105,8 @@ class S3FileSystem(FileSystemBase):
             with self._client.put_object(bucket, key) as stream:
                 yield stream
         elif mode == "rb":  # read mode
-            logger.debug("create_stream readable for %s", path_str)
-            relative_path = os.path.relpath(path, self._path)
-            if self.file_ranges and relative_path in self.file_ranges:
-                ranges = self.file_ranges[relative_path]
-                # ! Force use list_of_ranges reader for now
-                # TODO: (Important) improve this by passing in ranges parameter properly
-                reader_constructor = S3ReaderConstructor.list_of_ranges(ranges)
-
-            # Use provided reader_constructor or fall back to default
-            constructor = reader_constructor or self._reader_constructor
             with self._client.get_object(
-                bucket, key, reader_constructor=constructor
+                bucket, key, reader_constructor=self._reader_constructor
             ) as stream:
                 yield stream
         else:
@@ -380,20 +373,21 @@ class S3StorageReader(FileSystemReader):
             LoadPlan: The same plan with items sorted by storage offset.
         """
 
-        # Calculate ranges per file
-        per_file_ranges = {}
-        for read_item in plan.items:
-            item_md = self.storage_data[read_item.storage_index]
-            path = item_md.relative_path
-            if path not in per_file_ranges:
-                per_file_ranges[path] = []
-            per_file_ranges[path].append(
-                RangeRequest(start=item_md.offset, end=item_md.offset + item_md.length)
-            )
-
-        # Store ranges in filesystem
-        # TODO find a better place to handle this information
-        self.fs.file_ranges = per_file_ranges
+        # Inject ranges if using DCP list-of-ranges reader constructor
+        if isinstance(self.fs._reader_constructor, DCPListOfRangesConstructor):
+            # Calculate ranges per file
+            per_file_ranges = {}
+            for read_item in plan.items:
+                item_md = self.storage_data[read_item.storage_index]
+                path = item_md.relative_path
+                if path not in per_file_ranges:
+                    per_file_ranges[path] = []
+                per_file_ranges[path].append(
+                    RangeRequest(
+                        start=item_md.offset, end=item_md.offset + item_md.length
+                    )
+                )
+            self.fs._reader_constructor.set_file_ranges(per_file_ranges)
 
         # Sort items in plan based on their offset in checkpoints shards
         plan.items.sort(key=lambda item: self.storage_data[item.storage_index].offset)
--- a/s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
+++ b/s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
@@ -7,7 +7,7 @@ import os
 import urllib.parse
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Generator, Union, Optional
+from typing import Generator, Union, Optional, Dict
 from typing import List
 
 from s3torchconnectorclient._mountpoint_s3_client import S3Exception
@@ -56,6 +56,7 @@ class S3FileSystem(FileSystemBase):
         """
         self._path: Union[str, os.PathLike] = ""
         self._reader_constructor = reader_constructor or S3ReaderConstructor.default()
+        self.file_ranges: Optional[Dict[str, List[RangeRequest]]] = None
 
         # Get reader type string for user agent
         reader_type_string = S3ReaderConstructor.get_reader_type_string(
@@ -75,7 +76,10 @@ class S3FileSystem(FileSystemBase):
 
     @contextmanager
     def create_stream(
-        self, path: Union[str, os.PathLike], mode: str
+        self,
+        path: Union[str, os.PathLike],
+        mode: str,
+        reader_constructor: Optional[S3ReaderConstructorProtocol] = None,
     ) -> Generator[io.IOBase, None, None]:
         """
         Create a stream for reading or writing to S3.
@@ -99,8 +103,17 @@ class S3FileSystem(FileSystemBase):
                 yield stream
         elif mode == "rb":  # read mode
             logger.debug("create_stream readable for %s", path_str)
+            relative_path = os.path.relpath(path, self._path)
+            if self.file_ranges and relative_path in self.file_ranges:
+                ranges = self.file_ranges[relative_path]
+                # ! Force use list_of_ranges reader for now
+                # TODO: (Important) improve this by passing in ranges parameter properly
+                reader_constructor = S3ReaderConstructor.list_of_ranges(ranges)
+
+            # Use provided reader_constructor or fall back to default
+            constructor = reader_constructor or self._reader_constructor
             with self._client.get_object(
-                bucket, key, reader_constructor=self._reader_constructor
+                bucket, key, reader_constructor=constructor
             ) as stream:
                 yield stream
         else:
@@ -318,6 +331,17 @@ class S3StorageWriter(FileSystemWriter):
         return S3FileSystem.validate_checkpoint_id(checkpoint_id)
 
 
+import io
+from typing import Dict, List, Optional
+from dataclasses import dataclass
+
+import torch
+from torch.distributed.checkpoint.filesystem import FileSystemReader
+from torch.distributed.checkpoint.planner import LoadPlan
+
+from s3torchconnector.s3reader.list_of_ranges import RangeRequest
+
+
 class S3StorageReader(FileSystemReader):
     def __init__(
         self,
@@ -355,6 +379,22 @@ class S3StorageReader(FileSystemReader):
         Returns:
             LoadPlan: The same plan with items sorted by storage offset.
         """
+
+        # Calculate ranges per file
+        per_file_ranges = {}
+        for read_item in plan.items:
+            item_md = self.storage_data[read_item.storage_index]
+            path = item_md.relative_path
+            if path not in per_file_ranges:
+                per_file_ranges[path] = []
+            per_file_ranges[path].append(
+                RangeRequest(start=item_md.offset, end=item_md.offset + item_md.length)
+            )
+
+        # Store ranges in filesystem
+        # TODO find a better place to handle this information
+        self.fs.file_ranges = per_file_ranges
+
         # Sort items in plan based on their offset in checkpoints shards
         plan.items.sort(key=lambda item: self.storage_data[item.storage_index].offset)
         return plan
--- a/s3torchconnector/src/s3torchconnector/s3reader/constructor.py
+++ b/s3torchconnector/src/s3torchconnector/s3reader/constructor.py
@@ -2,12 +2,33 @@
 #  // SPDX-License-Identifier: BSD
 
 from functools import partial
-from typing import Optional, List
+from typing import Optional, List, Dict
 
+from .s3reader import S3Reader
 from .protocol import S3ReaderConstructorProtocol
 from .sequential import SequentialS3Reader
 from .ranged import RangedS3Reader
 from .list_of_ranges import ListOfRangesS3Reader, RangeRequest
+
+
+class DCPListOfRangesConstructor:
+    def __init__(self) -> None:
+        self._file_ranges: Dict[str, List[RangeRequest]] = {}
+
+    def set_file_ranges(self, file_ranges: Dict[str, List[RangeRequest]]) -> None:
+        self._file_ranges = file_ranges
+
+    def __call__(
+        self, bucket: str, key: str, get_object_info, get_stream, **kwargs
+    ) -> S3Reader:
+        if self._file_ranges:
+            # TODO: Check if using filename only works with prefix strategies
+            filename = key.split("/")[-1]
+            return ListOfRangesS3Reader(
+                bucket, key, self._file_ranges[filename], get_object_info, get_stream
+            )
+        # Fallback to SequentialS3Reader if no file_ranges yet (e.g. when reading .metadata)
+        return SequentialS3Reader(bucket, key, get_object_info, get_stream, **kwargs)
 
 
 class S3ReaderConstructor:
@@ -94,6 +115,12 @@ class S3ReaderConstructor:
         return partial(ListOfRangesS3Reader, ranges=ranges)
 
     @staticmethod
+    def dcp_list_of_ranges() -> S3ReaderConstructorProtocol:
+        """Creates a DCP-optimized constructor that uses ListOfRanges when ranges are available"""
+        # TODO update docstring
+        return DCPListOfRangesConstructor()
+
+    @staticmethod
     def default() -> S3ReaderConstructorProtocol:
         """Creates default reader constructor (sequential)
 
@@ -112,10 +139,11 @@ class S3ReaderConstructor:
                 S3ReaderConstructor.default()
             )
 
-        if not isinstance(constructor, partial):
+        if isinstance(constructor, DCPListOfRangesConstructor):
+            return "dcp_list_of_ranges"
+        elif not isinstance(constructor, partial):
             return "unknown"
-
-        if constructor.func == RangedS3Reader:
+        elif constructor.func == RangedS3Reader:
             return "range_based"
         elif constructor.func == SequentialS3Reader:
             return "sequential"
--- a/s3torchconnector/src/s3torchconnector/s3reader/constructor.py
+++ b/s3torchconnector/src/s3torchconnector/s3reader/constructor.py
@@ -2,11 +2,12 @@
 #  // SPDX-License-Identifier: BSD
 
 from functools import partial
-from typing import Optional
+from typing import Optional, List
 
 from .protocol import S3ReaderConstructorProtocol
 from .sequential import SequentialS3Reader
 from .ranged import RangedS3Reader
+from .list_of_ranges import ListOfRangesS3Reader, RangeRequest
 
 
 class S3ReaderConstructor:
@@ -24,7 +25,10 @@ class S3ReaderConstructor:
     """
 
     @staticmethod
-    def sequential() -> S3ReaderConstructorProtocol:
+    def sequential(
+        start_offset: Optional[int] = None,
+        end_offset: Optional[int] = None,
+    ) -> S3ReaderConstructorProtocol:
         """Creates a constructor for sequential readers
 
         Returns:
@@ -35,7 +39,12 @@ class S3ReaderConstructor:
             reader_constructor = S3ReaderConstructor.sequential()
 
         """
-        return partial(SequentialS3Reader)
+        # TODO update docstrings (after implementation fixed)
+        return partial(
+            SequentialS3Reader,
+            start_offset=start_offset,
+            end_offset=end_offset,
+        )
 
     @staticmethod
     def range_based(buffer_size: Optional[int] = None) -> S3ReaderConstructorProtocol:
@@ -79,6 +88,12 @@ class S3ReaderConstructor:
         return partial(RangedS3Reader, buffer_size=buffer_size)
 
     @staticmethod
+    def list_of_ranges(ranges: List[RangeRequest]) -> S3ReaderConstructorProtocol:
+        """Creates a constructor for ListOfRangesS3Reader with specific ranges"""
+        # TODO update docstring
+        return partial(ListOfRangesS3Reader, ranges=ranges)
+
+    @staticmethod
     def default() -> S3ReaderConstructorProtocol:
         """Creates default reader constructor (sequential)
 
@@ -104,5 +119,7 @@ class S3ReaderConstructor:
             return "range_based"
         elif constructor.func == SequentialS3Reader:
             return "sequential"
+        elif constructor.func == ListOfRangesS3Reader:
+            return "list_of_ranges"
         else:
             return "unknown"
--- a/s3torchconnector/src/s3torchconnector/s3reader/__init__.py
+++ b/s3torchconnector/src/s3torchconnector/s3reader/__init__.py
@@ -2,9 +2,10 @@
 #  // SPDX-License-Identifier: BSD
 
 from .s3reader import S3Reader
-from .constructor import S3ReaderConstructor
+from .constructor import S3ReaderConstructor, DCPListOfRangesConstructor
 from .sequential import SequentialS3Reader
 from .ranged import RangedS3Reader
+from .list_of_ranges import ListOfRangesS3Reader
 from .protocol import GetStreamCallable, S3ReaderConstructorProtocol
 
 __all__ = [
@@ -12,4 +13,5 @@ __all__ = [
     "S3ReaderConstructor",
     "SequentialS3Reader",
     "RangedS3Reader",
+    "ListOfRangesS3Reader",
 ]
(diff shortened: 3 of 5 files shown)


s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,358 @@
...
+        for i in range(self._current_group_idx, len(self._range_groups)):
+            group = self._range_groups[i]
+            if group.start <= item.start <= item.end <= group.end:
+                # Create new stream if switching groups, or no stream exists (for group 0)
+                if self._stream is None or i != self._current_group_idx:
+                    self._current_group_idx = i
+                    self._stream = self._get_stream(group.start, group.end)
+                    self._stream_pos = group.start
+                    self._leftover = b""
+                return
```
jet-tong: Similarly from forward search above - considered checking current, and then next item, and then raise error. We do support skipping a full group, but it would be hard to skip a full group without skipping some individual ItemRanges.

s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
```diff
@@ -56,6 +60,7 @@ def __init__(
         """
         self._path: Union[str, os.PathLike] = ""
         self._reader_constructor = reader_constructor or S3ReaderConstructor.default()
+        self.file_ranges: Optional[Dict[str, List[ItemRange]]] = None
```
muddyfish: This doesn't appear to be used

s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
```diff
@@ -355,8 +377,23 @@ def prepare_local_plan(self, plan: LoadPlan) -> LoadPlan:
         Returns:
             LoadPlan: The same plan with items sorted by storage offset.
         """
+
         # Sort items in plan based on their offset in checkpoints shards
         plan.items.sort(key=lambda item: self.storage_data[item.storage_index].offset)
+
+        # Inject ranges if using DCP optimized reader constructor
+        if isinstance(self.fs._reader_constructor, DCPOptimizedConstructor):
```
muddyfish: This is janky - why are we doing things to the reader constructor here? Is there a method we should add to `S3ReaderConstructorProtocol` that does the local plan preparation?

muddyfish: Otherwise, should we create a new protocol that knows how to do local plan checking?

s3torchconnector/src/s3torchconnector/s3reader/constructor.py
```diff
@@ -97,10 +131,11 @@ def get_reader_type_string(
                 S3ReaderConstructor.default()
             )
 
-        if not isinstance(constructor, partial):
+        if isinstance(constructor, DCPOptimizedConstructor):
```
muddyfish: Same here - this feels pretty janky to me. What's this used for? Just debugging or to actually do something based on it?

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def _validate_and_coalesce_ranges(
+        self, ranges: List[ItemRange], max_gap_size: int
+    ) -> List[RangeGroup]:
+        """Coalescing nearby byte ranges within max_gap_size."""
+        if not ranges:
+            return []
+
+        groups: List[RangeGroup] = []
+        current = [ranges[0]]
```
muddyfish: These variables could be better - what type is this? It's the WIP `RangeGroup`, not an `ItemRange`

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def _get_stream_for_item(self, item: ItemRange) -> None:
+        """Find which RangeGroup contains the given position."""
+
+        # Forward search from current RangeGroup (should always be current or next group)
+        for i in range(self._current_group_idx, len(self._range_groups)):
+            group = self._range_groups[i]
+            if group.start <= item.start <= item.end <= group.end:
+                # Create new stream if switching groups, or no stream exists (for group 0)
+                if self._stream is None or i != self._current_group_idx:
```
muddyfish: Should this return a stream object?

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,358 @@
...
+            self._current_item_idx = item_idx
+            return
+
+        # Get stream from the right RangeGroup for start_pos
+        self._get_stream_for_item(item)
+        assert self._stream is not None
+
+        current_pos = self._stream_pos
+        buffer = self._leftover
+        remaining = item.end - item.start
```
muddyfish: Why is this called `remaining` when it's the length of the item?

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,358 @@
...
+        chunks: List[bytes] = []
+
+        # TODO: check BytesIO.write() vs b"".join() + BytesIO(data)
+
+        # 1. Serve from buffered leftover bytes
+        if buffer and current_pos <= item.start < current_pos + len(buffer):
+            offset = item.start - current_pos
+            take = min(remaining, len(buffer) - offset)
```
muddyfish: This logic is slightly confusing - can we simplify?

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,358 @@
...
+        return self._position
+
+    def tell(self) -> int:
+        """
+        Returns:
+            int: Current absolute position in the object.
+        """
+        return self._position
+
+    def close(self) -> None:
```
muddyfish: Should set `_closed`

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,358 @@
...
+            return b""
+
+        item_idx = self._find_item_for_position(self._position)
+        if item_idx != self._current_item_idx or self._current_item_buffer is None:
+            self._load_item_buffer(item_idx)
+
+        item_range = self._item_ranges[self._current_item_idx]
+        local_pos = self._position - item_range.start
+
+        assert self._current_item_buffer is not None
```
muddyfish: We should check that the read doesn't take us outside of the current item

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,358 @@
...
+
+        Returns:
+            int: Current position of the stream
+
+        Raises:
+            TypeError: If whence is not SEEK_SET or SEEK_CUR.
+            ValueError: If seeking to negative position or accessing previous items.
+            TypeError: If whence is not SEEK_SET or SEEK_CUR.
+        """
+        if not isinstance(offset, int):
```
muddyfish: There's no bounds checking here to make sure we stay within the current item

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
jet-tong: for tracking remaining bytes - can update name / logic here.

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,358 @@
...
+                chunks.append(chunk[:remaining])
+                buffer = chunk[remaining:]
+                current_pos += remaining
+                remaining = 0
+                break
+
+        self._stream_pos = current_pos
+        self._leftover = buffer
+        data = b"".join(chunks)
+        self._current_item_buffer = io.BytesIO(data)
```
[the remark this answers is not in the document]
jet-tong: We have to store it in self._current_item_buffer for multiple read() or readinto() calls to use later.

Perhaps it's better to return the buffer, and then apply the store in read e.g.

```python
        if item_idx != self._current_item_idx or self._current_item_buffer is None:
            self._current_item_buffer = self._load_item_buffer(item_idx)
```

s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
jet-tong: > new protocol that knows how to do local plan checking

Done. Logic also moved to DCPOptimizedConstructor. 

Can reopen this discussion if there is a better path for injecting ranges.

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
jet-tong: Changing to `items: List[ItemRange]` (current group's ItemRanges)

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,374 @@
...
+        if leftover and pos <= item.start < pos + len_leftover:
+            start = item.start - pos
+            available_bytes = len_leftover - start
+            size = min(bytes_left, available_bytes)
+            end = start + size
+
+            chunks.append(leftover[start:end])
+            bytes_left -= size
+            pos = item.start + size
+            leftover = leftover[end:] if end < len_leftover else b""
```
jet-tong: Bug here: if `item.start > pos + len_leftover`, we didn't advance `pos` and clear `leftover`. 

This would fix:
```python
        elif leftover and item.start >= pos + len_leftover:
            # Target beyond leftover: advance pos
            pos += len_leftover
            leftover = b""
```

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,358 @@
...
+        item_idx = self._find_item_for_position(self._position)
+        if item_idx != self._current_item_idx or self._current_item_buffer is None:
+            self._load_item_buffer(item_idx)
+
+        item_range = self._item_ranges[self._current_item_idx]
+        local_pos = self._position - item_range.start
+
+        assert self._current_item_buffer is not None
+        self._current_item_buffer.seek(local_pos)
+        bytes_read = self._current_item_buffer.readinto(buf)
```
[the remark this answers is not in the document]
jet-tong: Now delegated to `bytes_to_read = min(dest_len, size - pos)` in `_ItemViewBuffer`

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,563 @@
...
+            raise ValueError(
+                f"{FIND_ITEM_ERROR_PREFIX}Position {pos} before current range "
+                f"{self._current_item.start}-{self._current_item.end}"
+            )
+
+        # Return item if position still in current item
+        if pos < self._current_item.end:
+            return self._current_item
+
+        # Iterate through remaining items
```
jet-tong: Iterate next item - not remaining items

s3torchconnector/pyproject.toml
```diff
@@ -34,7 +34,8 @@ test = [
     "hypothesis",
     "flake8",
     "black",
-    "mypy"
+    "mypy",
+    "importlib_metadata; python_version == '3.9'", # PyTorch 2.7.0+ DCP w/ Python 3.9 requires this module; for dcp_optimized reader unit tests
```
muddyfish: Only for 3.9? Is this the earliest Python version we support now?

s3torchconnector/src/s3torchconnector/s3reader/constructor.py
```diff
@@ -1,12 +1,71 @@
...
+        self._max_gap_size: Union[int, float] = max_gap_size
+
+    def set_item_ranges_by_file(
+        self,
+        plan_items: "List[ReadItem]",
+        storage_data: "Dict[MetadataIndex, _StorageInfo]",
+    ) -> None:
+
+        if not plan_items:
+            return  # Allow lack of plan_items, for SequentialS3Reader fallbacks
```
muddyfish: Is this correct? Did we decide what we wanted to do in case this method was called multiple times?

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,588 @@
...
+
+
+# TODO: extend buffer for use in other S3Reader implementations after extensive testing
+class _ItemViewBuffer:
+    """
+    A tiny, zero-copy, read-only buffer built from multiple memoryview segments.
+    Replaces io.BytesIO which involved extra copies for creation and buffer growth.
+    """
+
+    __slots__ = ("_segments", "_offsets", "_lengths", "_size", "_pos", "_closed")
```
muddyfish: Why did we choose to implement `__slots__` here?

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,588 @@
...
+    def __init__(self) -> None:
+        self._segments: List[memoryview] = []  # memoryview segments
+        self._offsets: List[int] = []  # start offset (within the item) of each segment
+        self._lengths: List[int] = []  # length of each segment
+        self._size: int = 0  # total item length (sum of _lengths)
+        self._pos: int = 0  # current read position within the item
+        self._closed: bool = False
+
+    def append_view(self, view: memoryview) -> None:
+        """Append a memoryview segment (ignored if empty)."""
```
muddyfish: Is it valid to append more views whilst the position is non-zero?

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,588 @@
...
+    """
+    A tiny, zero-copy, read-only buffer built from multiple memoryview segments.
+    Replaces io.BytesIO which involved extra copies for creation and buffer growth.
+    """
+
+    __slots__ = ("_segments", "_offsets", "_lengths", "_size", "_pos", "_closed")
+
+    def __init__(self) -> None:
+        self._segments: List[memoryview] = []  # memoryview segments
+        self._offsets: List[int] = []  # start offset (within the item) of each segment
```
muddyfish: Is this effectively just the cumsum of `self._lengths`?

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,588 @@
...
+        # Seeking past EOF is allowed.
+        self._pos = new_pos
+        return self._pos
+
+    def tell(self) -> int:
+        """Return the current pos position (like BytesIO.tell)."""
+        return self._pos
+
+    def read(self, size: Optional[int] = None) -> bytes:
+        assert size is not None, "Size cannot be None; full read is not supported"
```
muddyfish: Why not?

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,588 @@
...
+        if size == 4 and self._pos == 0 and self._lengths and self._lengths[0] >= 4:
+            self._pos = 4
+            # TODO: eliminating bytes() conversion can save ~3% time? Requires interface changes.
+            return bytes(self._segments[0][:4])
+
+        if size == 0:
+            return b""
+
+        # Pass implementation to readinto()
+        out = bytearray(size)
```
muddyfish: Could coerce to `memoryview` here and then wouldn't have to do the fancy ternary later

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,588 @@
...
+        assert not dest.readonly, "writable buffer required"
+
+        dest_len = len(dest)
+        size = self._size
+        pos = self._pos
+
+        if dest_len == 0 or pos >= size:
+            return 0
+
+        # Cache to avoid repeated attribute calls
```
muddyfish: This is very low level optimisation. Given you're using `__slots__`, doesn't that typically remove the cost of `self.` getattr calls?

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,588 @@
...
+    def _get_stream_for_item(self, item: ItemRange) -> GetObjectStream:
+        """Find which RangeGroup contains the given position."""
+
+        # If item is the first item of a new group, create new stream
+        if item.start in self._group_start_to_group:
+            group = self._group_start_to_group[item.start]
+            self._stream = self._get_stream(group.start, group.end)
+            self._stream_pos = group.start
+            self._leftover = None
+            return self._stream
```
muddyfish: I feel there should be a subclass/dataclass just for handling the attributes on the stream

s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py
```diff
@@ -337,7 +347,12 @@ def __init__(
                 e.g. S3ReaderConstructor.sequential() or S3ReaderConstructor.range_based()
         """
         super().__init__(path)
-        self.fs = S3FileSystem(region, s3client_config=s3client_config, reader_constructor=reader_constructor)  # type: ignore
+        self._reader_constructor = reader_constructor or S3ReaderConstructor.default()
+        self.fs: S3FileSystem = S3FileSystem(  # type: ignore[assignment]
```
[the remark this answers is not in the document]
jet-tong: Type ignore was in original code. mypy errors:
```
# with type hint - self.fs: S3FileSystem = S3FileSystem(...
s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py:351: error: Incompatible types in assignment (expression has type "S3FileSystem", base class "FileSystemReader" defined the type as "FileSystem")  [assignment]
# without type hint - self.fs = S3FileSystem(...
s3torchconnector/src/s3torchconnector/dcp/s3_file_system.py:351: error: Incompatible types in assignment (expression has type "S3FileSystem", variable has type "FileSystem")  [assignment]
```
[Reference code](https://github.com/pytorch/pytorch/blob/main/torch/distributed/checkpoint/_fsspec_filesystem.py#L1-L2) has file-wide mypy ignore-errors.

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
jet-tong: It should have slightly faster attribute lookup speeds - an optimization I found since this buffer class's readinto() method will be a hotpath for optimizing for loading speeds.

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
jet-tong: That's correct - separate variables to avoid runtime calc.

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
jet-tong: Yes (if I understood correctly)

I added this method to mimic `BytesIO.write(chunk)`, and instead we can do `_ItemViewBuffer.append_view(memoryview(chunk))`.

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,588 @@
...
+        self._lengths.append(seg_len)
+        self._size += seg_len
+
+    def close(self) -> None:
+        if not self._closed:
+            self._closed = True
+            self._segments.clear()
+
+    def seek(self, offset: int, whence: int = SEEK_SET, /) -> int:
+        assert isinstance(offset, int), f"integer expected, got {type(offset)!r}"
```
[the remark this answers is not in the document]
jet-tong: As per suggestion above: Get rid of close() entirely in _ItemViewBuffer, and allow garbage collector to deallocate resources. 

Asserting not closed() might require extra time used so I hasn't implemented although it's not best practice - but if we get rid of close this doesn't matter as much.

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
jet-tong: True, we can. 

I kept this ternary (`bytes(out) if n == size else memoryview(out)[:n].tobytes()`) before while they were bytes and kept it - which reduces number of slicing operations by a lot. (Slicing should only happen when incorrect reads happen). 

But now that we use memoryviews I guess it's tiny tradeoff between an if check, and a memoryview slice operation (for using vs not using ternary) - not sure if I should simply use `memoryview(out)[:n].tobytes()` for all.

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,588 @@
...
+        get_stream: Callable[[Optional[int], Optional[int]], GetObjectStream],
+        max_gap_size: Union[int, float] = DEFAULT_MAX_GAP_SIZE,
+    ):
+        if not bucket:
+            raise ValueError("Bucket should be specified")
+        if not key:
+            raise ValueError("Key should be specified")
+        if not item_ranges:
+            raise ValueError("item_ranges must be a non-empty List[ItemRange] object")
+        if not isinstance(max_gap_size, (int, float)):
```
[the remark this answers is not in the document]
jet-tong: Decided to allow float("inf"), I assume is the more common way to convey maxsize, for coalescing all ItemRanges into a single stream.

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,588 @@
...
+
+        # Otherwise, we're still in same group - reuse stream created when reading 1st item
+        if self._stream is None:
+            raise ValueError(
+                f"{FIND_ITEM_ERROR_PREFIX}Attempted to read item {item.start}-{item.end} "
+                f"without starting at the first item of its range-group"
+            )
+        return self._stream
+
+    def _get_item_buffer(self, item: ItemRange) -> _ItemViewBuffer:
```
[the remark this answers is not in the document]
jet-tong: Good suggestion! Will have a look.

s3torchconnector/tst/unit/test_s3reader_constructor.py
```diff
@@ -38,32 +61,262 @@ def test_s3readerconstructor_range_based_constructor_buffer_configurations(
...
+    assert constructor._max_gap_size == max_gap_size
+
+    constructor._item_ranges_by_file = {TEST_PATH: [ItemRange(0, 100)]}
+    s3reader = constructor(TEST_BUCKET, TEST_KEY, MOCK_OBJECT_INFO, MOCK_STREAM)
+    assert isinstance(s3reader, DCPOptimizedS3Reader)
+    assert s3reader._max_gap_size == max_gap_size
+
+
+@pytest.mark.parametrize("max_gap_size", [sys.maxsize, float("inf"), 0.5])
+def test_dcp_optimized_constructor_max_gap_size_edge_cases(max_gap_size):
```
muddyfish: Why is this a different test from above?

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,737 @@
...
+        # Cache to avoid repeated attribute lookups in the loop
+        segments = self._segments
+        offsets = self._offsets
+        lengths = self._lengths
+
+        # Find starting segment using binary search: last i where offsets[i] <= pos
+        # Using bisect (not caching) since torch.load jumps around (magic bytes, zip dir, tensor data)
+        seg_idx = bisect.bisect_right(offsets, pos) - 1
+        if seg_idx < 0:
+            seg_idx = 0
```
muddyfish: Not obvious why you need special logic to make 0 and -1 act the same.

jet-tong: Will add comment, thanks.

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,737 @@
...
+
+        # If item is the first item of a new group, create new stream
+        if item.start in self._group_start_to_group:
+            group = self._group_start_to_group[item.start]
+            self._stream_state.stream = self._get_stream(group.start, group.end)
+            self._stream_state.stream_position = group.start
+            self._stream_state.leftover = None
+            return self._stream_state.stream
+
+        # Otherwise, we're still in same group - reuse stream created when reading 1st item
```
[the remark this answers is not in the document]
jet-tong: Sequential access is already enforced in _find_item_for_position, which runs  _get_item_buffer (which calls this _get_stream_for_item method). Will add comment for that, and can discuss if more changes needed to make it clearer. Won't add any new asserts since most are addressed in _find_item_for_position.

s3torchconnector/src/s3torchconnector/s3reader/dcp_optimized.py
```diff
@@ -0,0 +1,751 @@
...
+        # Cache to avoid repeated attribute lookups in the loop
+        segments = self._segments
+        offsets = self._offsets
+        lengths = self._lengths
+
+        # Find segment containing pos with binary search
+        # bisect_right gives insertion point where pos < offsets[i], -1 gives containing segment.
+        # No caching optimisation, since torch.load jumps around (magic bytes, zip dir, tensor data)
+        seg_idx = bisect.bisect_right(offsets, pos) - 1
+        # Defensive clamp: shouldn't occur as _find_item_for_range makes sure pos >= 0
```
muddyfish: What are the consequences of this happening? Should we raise an assertion error?

s3torchconnector/tst/unit/test_s3reader_dcp_optimized.py
```diff
@@ -0,0 +1,1229 @@
...
+        """Reader position == reference position after synced operations (not failure cases)."""
+        assert self.reader.tell() >= self.reference_io.tell()
+
+    # -------------------------------------------------------------------------
+    # Rules: Success cases - seek and read within current/next item
+    # -------------------------------------------------------------------------
+
+    @precondition(lambda self: self._has_current_item())
+    @rule(data=data(), use_readinto=booleans())
+    def seek_and_read_within_current_item(self, data, use_readinto):
```
muddyfish: There should be another rule for continuing to read without seeking

s3torchconnector/tst/unit/test_s3reader_dcp_optimized.py
```diff
@@ -0,0 +1,1229 @@
...
+    # -------------------------------------------------------------------------
+
+    @precondition(lambda self: self._has_current_item())
+    @rule(data=data(), use_readinto=booleans())
+    def seek_current_read_beyond(self, data, use_readinto):
+        """Failure 1: Seek within current item, but read extends beyond it."""
+        note(f"CASE FAILURE 1: Seek within current item, but read extends beyond it")
+
+        item = self._current_item()
+        saved_state = self._save_reader_state()
```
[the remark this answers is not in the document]
jet-tong: Excellent suggestion! Used context manager for this one now and it's now a helper function.

PR #420, opened 2026-02-26
Add S3 endpoint support to s3torchbenchmarking

ovaistariq:
## Description

Add support for custom S3-compatible endpoints to the s3torchbenchmarking tool. The underlying s3torchconnector library already supports custom endpoints via an `endpoint` parameter, but the benchmarking tool never passed this configuration through. This change enables using the benchmarking suite against any S3-compatible storage service.

## Additional context

No breaking changes. The endpoint parameter is optional and defaults to None throughout, preserving existing AWS S3 behavior.

## Testing

The existing smoke test (module imports) passes.

--------

### DIFF
--- a/s3torchbenchmarking/conf/dataset.yaml
+++ b/s3torchbenchmarking/conf/dataset.yaml
@@ -8,6 +8,7 @@ defaults:
 s3:
   region: ??? # e.g., eu-west-1
   bucket: ??? # e.g., my-bucket (*not* an S3 URI)
+  endpoint: null # optional: custom S3-compatible endpoint URL
 # Boolean flag to tell whether the dataset is sharded or not.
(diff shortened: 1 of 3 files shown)


Mattral (commented):
Hello there, trying to learn something from your PR. Thanks for adding support for custom endpoints and path‑style addressing, this will make benchmarking against non‑AWS S3 services much easier. One question: since different S3‑compatible providers vary in how they handle authentication and addressing, have you considered documenting or testing against at least one non‑AWS provider (e.g., MinIO or Ceph) to validate that the endpoint + force_path_style flags behave consistently? I believe that would help maintainers ensure forward compatibility and give contributors confidence when extending support to other backends.

PR #428, opened 2026-06-08
fix(config): validate max_attempts in S3ClientConfig to avoid opaque Rust panic

### LINKED ISSUE #361: Setting `max_attempts>64` results in PanicException
### s3torchconnector version

1.4.2

### s3torchconnectorclient version

1.4.2

### AWS Region

us-east-2

### Describe the running environment

Running in EC2 p5e

### What happened?

Setting

```python
S3ClientConfig(
    max_attempts=65,
)
```

Results in the error (with `RUST_BACKTRACE=full`)

```python
E             File "/home/ubuntu/carlos/titan/.venv/lib/python3.10/site-packages/s3torchconnector/dcp/s3_file_system.py", line 192, in exists
E               self._client.head_object(bucket, key)
E             File "/home/ubuntu/carlos/titan/.venv/lib/python3.10/site-packages/s3torchconnector/_s3client/_s3client.py", line 197, in head_object
E               return self._client.head_object(bucket, key)
E             File "/home/ubuntu/carlos/titan/.venv/lib/python3.10/site-packages/s3torchconnector/_s3client/_s3client.py", line 116, in _client
E               self._native_client = self._client_builder()
E             File "/home/ubuntu/carlos/titan/.venv/lib/python3.10/site-packages/s3torchconnector/_s3client/_s3client.py", line 137, in _client_builder
E               return MountpointS3Client(
E           pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: Error(34, "aws-c-common: AWS_ERROR_INVALID_ARGUMENT, An invalid argument was passed to a function.")

.venv/lib/python3.10/site-packages/torch/distributed/checkpoint/utils.py:219: CheckpointException
------------------------------------------------------------- Captured stderr call --------------------------------------------------------------
thread '<unnamed>' panicked at /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/mountpoint-s3-client-0.14.1/src/s3_crt_client.rs:355:74:
called `Result::unwrap()` on an `Err` value: Error(34, "aws-c-common: AWS_ERROR_INVALID_ARGUMENT, An invalid argument was passed to a function.")
stack backtrace:
   0:     0x7fa19e19e64a - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h14b23c1989cbd5c2
   1:     0x7fa19e1c0ad3 - core::fmt::write::h346b5eee5ed4d7cc
   2:     0x7fa19e19c0c3 - std::io::Write::write_fmt::heeb48dcd4a60b46b
   3:     0x7fa19e19e492 - std::sys::backtrace::BacktraceLock::print::h48db11f3fd4983ff
   4:     0x7fa19e19f45c - std::panicking::default_hook::{{closure}}::h65db976b9c0d8674
   5:     0x7fa19e19f2a2 - std::panicking::default_hook::h6eb3a1192db1ae36
   6:     0x7fa19e19fa37 - std::panicking::rust_panic_with_hook::h370ee1901241e459
   7:     0x7fa19e19f8ca - std::panicking::begin_panic_handler::{{closure}}::heabfe92676d6d073
   8:     0x7fa19e19eb29 - std::sys::backtrace::__rust_end_short_backtrace::h6e22d229d4fdf49e
   9:     0x7fa19e19f55c - rust_begin_unwind
  10:     0x7fa19dc39280 - core::panicking::panic_fmt::hfae270fab21da3e6
  11:     0x7fa19dc39666 - core::result::unwrap_failed::h82c78b552479728a
  12:     0x7fa19dcd9901 - mountpoint_s3_client::s3_crt_client::S3CrtClient::new::ha8f4992ede3e0af4
  13:     0x7fa19dc3b0da - _mountpoint_s3_client::mountpoint_s3_client::MountpointS3Client::new_s3_client::h3eb7ef726f1a536b
  14:     0x7fa19dc3d097 - _mountpoint_s3_client::mountpoint_s3_client::<impl pyo3::impl_::pyclass::PyMethods<_mountpoint_s3_client::mountpoint_s3_client::MountpointS3Client> for pyo3::impl_::pyclass::PyClassImplCollector<_mountpoint_s3_client::mountpoint_s3_client::MountpointS3Client>>::py_methods::ITEMS::trampoline::ha316fa5b6e2cd108
  15:     0x7fa59eceff54 - type_call
                               at /build/Python-3.10.12/Objects/typeobject.c:1123:11
  16:     0x7fa59ed3c8d6 - _PyObject_MakeTpCall
                               at /build/Python-3.10.12/Objects/call.c:215:18
  17:     0x7fa59ed3c8d6 - _PyObject_VectorcallTstate
```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

valter-silva-au:
## Problem

`S3ClientConfig(max_attempts=N)` with an out-of-range `N` crashes with an opaque `pyo3_runtime.PanicException` raised from inside the Rust client constructor. Because the panic crosses the PyO3/FFI boundary, it is **not catchable as a normal Python exception**, so callers cannot defend against it cleanly.

- `max_attempts=0` panics via Rust `NonZeroUsize::try_from(...).expect("max_attempts must be > 0")`.
- `max_attempts` above the AWS CRT exponential-backoff retry-strategy cap panics inside the CRT.

## Root cause

`s3torchconnector/_s3client/s3client_config.py` defines `S3ClientConfig` as a `@dataclass(frozen=True)` with **no validation**. Invalid values fall straight through the pure-Python layer into the Rust constructor, where they panic.

## Fix

Add a `__post_init__` guard that validates `max_attempts` at construction time and raises a clear, catchable `ValueError` instead of letting it reach the Rust panic:

- `max_attempts < 1` → `ValueError("max_attempts must be a positive integer")`
- `max_attempts > _MAX_ATTEMPTS_LIMIT` → `ValueError("max_attempts must be between 1 and 63 (AWS CRT retry-strategy limit)")`

**⚠️ Maintainer confirmation requested on the upper bound:** the exact cap is a CRT-internal constant. I've set `_MAX_ATTEMPTS_LIMIT = 63` as a single, clearly-commented module constant so it's trivial to adjust. **If you'd prefer to be conservative, I'm happy to ship only the lower-bound guard** (`< 1`, zero risk since the Rust side already rejects 0) and drop the upper bound — just let me know.

## Testing

Added to `tst/unit/test_s3_client_config.py`: a hypothesis property test `@given(integers(min_value=1, max_value=63))` confirming valid values construct fine, plus `test_max_attempts_zero_raises`, `test_max_attempts_negative_raises`, and `test_max_attempts_above_cap_raises`. Validated TDD-style (fail before, pass after).

## Risk

Low. Pure-Python, additive validation in a frozen dataclass (reads fields only). The lower-bound guard is zero-risk; the only judgment call is the exact upper-bound value, isolated to one adjustable constant.

Fixes #361

### DIFF
--- a/s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
+++ b/s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
@@ -3,6 +3,15 @@
 from dataclasses import dataclass
 from typing import Optional
 
+# Upper bound for `max_attempts`, enforced Python-side to avoid an opaque
+# pyo3_runtime.PanicException raised from inside the Rust client constructor
+# (see issue #361). The AWS CRT retry strategy uses exponential backoff and
+# rejects attempt counts above this limit. The exact value is a CRT-internal
+# constant; 63 matches the value cited by the maintainer. It is centralized
+# here so it can be adjusted in one place if the CRT limit is confirmed to
+# differ.
+_MAX_ATTEMPTS_LIMIT = 63
+
 
 @dataclass(frozen=True)
 class S3ClientConfig:
@@ -28,3 +37,18 @@ class S3ClientConfig:
     force_path_style: bool = False
     max_attempts: int = 10
     profile: Optional[str] = None
+
+    def __post_init__(self) -> None:
+        # Validate `max_attempts` here so that invalid values raise a clear,
+        # catchable ValueError instead of an opaque pyo3_runtime.PanicException
+        # from inside the Rust client constructor (see issue #361). The Rust
+        # side calls NonZeroUsize::try_from(max_attempts).expect(...) (panics on
+        # 0) and the AWS CRT retry strategy panics on values above its
+        # exponential-backoff limit.
+        if self.max_attempts < 1:
+            raise ValueError("max_attempts must be a positive integer")
+        if self.max_attempts > _MAX_ATTEMPTS_LIMIT:
+            raise ValueError(
+                f"max_attempts must be between 1 and {_MAX_ATTEMPTS_LIMIT} "
+                "(AWS CRT retry-strategy limit)"
+            )
--- a/s3torchconnector/tst/unit/test_s3_client_config.py
+++ b/s3torchconnector/tst/unit/test_s3_client_config.py
@@ -1,9 +1,11 @@
 #  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #  // SPDX-License-Identifier: BSD
+import pytest
 from hypothesis import given, example
 from hypothesis.strategies import integers, floats
 
 from s3torchconnector import S3ClientConfig
+from s3torchconnector._s3client.s3client_config import _MAX_ATTEMPTS_LIMIT
 from .test_s3_client import MiB, GiB
 
 
@@ -44,6 +46,32 @@ def test_throughput_target_gbps_setup(throughput_target_gbps: float):
 def test_max_attempts_setup(max_attempts: int):
     config = S3ClientConfig(max_attempts=max_attempts)
     assert config.max_attempts == max_attempts
+
+
+@given(max_attempts=integers(min_value=1, max_value=_MAX_ATTEMPTS_LIMIT))
+def test_max_attempts_valid_range(max_attempts: int):
+    config = S3ClientConfig(max_attempts=max_attempts)
+    assert config.max_attempts == max_attempts
+
+
+def test_max_attempts_zero_raises():
+    # See issue #361: max_attempts=0 panics inside the Rust constructor
+    # (NonZeroUsize::try_from(...).expect("max_attempts must be > 0")) as an
+    # uncatchable pyo3_runtime.PanicException. Guard it in pure Python instead.
+    with pytest.raises(ValueError):
+        S3ClientConfig(max_attempts=0)
+
+
+def test_max_attempts_negative_raises():
+    with pytest.raises(ValueError):
+        S3ClientConfig(max_attempts=-1)
+
+
+def test_max_attempts_above_cap_raises():
+    # See issue #361: values above the AWS CRT exponential-backoff retry-strategy
+    # limit panic inside the Rust constructor; reject them clearly in Python.
+    with pytest.raises(ValueError):
+        S3ClientConfig(max_attempts=_MAX_ATTEMPTS_LIMIT + 1)
 
 
 @given(


s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
```diff
@@ -28,3 +37,18 @@ class S3ClientConfig:
...
     max_attempts: int = 10
     profile: Optional[str] = None
+
+    def __post_init__(self) -> None:
+        # Validate `max_attempts` here so that invalid values raise a clear,
+        # catchable ValueError instead of an opaque pyo3_runtime.PanicException
+        # from inside the Rust client constructor (see issue #361). The Rust
+        # side calls NonZeroUsize::try_from(max_attempts).expect(...) (panics on
+        # 0) and the AWS CRT retry strategy panics on values above its
+        # exponential-backoff limit.
```
jet-tong, line 47: Let's remove this comment block - if we remove _MAX_ATTEMPTS_LIMIT and fix the 0 panic in s3torchconnectorclient, `self.max_attempts < 1:` with the value error raised should be self-explanatory.

s3torchconnector/tst/unit/test_s3_client_config.py
```diff
@@ -46,6 +48,32 @@ def test_max_attempts_setup(max_attempts: int):
...
+@given(max_attempts=integers(min_value=1, max_value=_MAX_ATTEMPTS_LIMIT))
+def test_max_attempts_valid_range(max_attempts: int):
+    config = S3ClientConfig(max_attempts=max_attempts)
+    assert config.max_attempts == max_attempts
+
+
+def test_max_attempts_zero_raises():
+    # See issue #361: max_attempts=0 panics inside the Rust constructor
+    # (NonZeroUsize::try_from(...).expect("max_attempts must be > 0")) as an
+    # uncatchable pyo3_runtime.PanicException. Guard it in pure Python instead.
```
jet-tong, line 60: Similarly, let's remove the comment block or reduce the verbosity if we can fix the Rust constructor directly in this PR.

s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
```diff
@@ -3,6 +3,15 @@
...
 from typing import Optional
 
+# Upper bound for `max_attempts`, enforced Python-side to avoid an opaque
+# pyo3_runtime.PanicException raised from inside the Rust client constructor
+# (see issue #361). The AWS CRT retry strategy uses exponential backoff and
+# rejects attempt counts above this limit. The exact value is a CRT-internal
+# constant; 63 matches the value cited by the maintainer. It is centralized
+# here so it can be adjusted in one place if the CRT limit is confirmed to
+# differ.
+_MAX_ATTEMPTS_LIMIT = 63
```
jet-tong, line 13: As discussed on my main review comment - let's ship only the lower-bound guard to be conservative as you suggested. 

Alternatively we could modify the docstring and softly document this limit (see below)
(Noting max attempt upper limit should be 64.)

s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
jet-tong, line 30: As an alternative, we may want to softly document the limits in the docstring, e.g.
```python
    max_attempts(int): amount of retry attempts for retryable errors. Must be greater than zero. Large values may be rejected by the underlying AWS CRT retry strategy (e.g. max_attempts>=65 in issue #361).
```

s3torchconnector/tst/unit/test_s3_client_config.py
```diff
@@ -46,6 +48,32 @@ def test_max_attempts_setup(max_attempts: int):
...
+def test_max_attempts_negative_raises():
+    with pytest.raises(ValueError):
+        S3ClientConfig(max_attempts=-1)
+
+
+def test_max_attempts_above_cap_raises():
+    # See issue #361: values above the AWS CRT exponential-backoff retry-strategy
+    # limit panic inside the Rust constructor; reject them clearly in Python.
+    with pytest.raises(ValueError):
+        S3ClientConfig(max_attempts=_MAX_ATTEMPTS_LIMIT + 1)
```
jet-tong, line 74: As discussed, let's remove it if we remove the upper limit guard.

s3torchconnector/tst/unit/test_s3_client_config.py
```diff
@@ -46,6 +48,32 @@ def test_max_attempts_setup(max_attempts: int):
     assert config.max_attempts == max_attempts
 
 
+@given(max_attempts=integers(min_value=1, max_value=_MAX_ATTEMPTS_LIMIT))
+def test_max_attempts_valid_range(max_attempts: int):
+    config = S3ClientConfig(max_attempts=max_attempts)
+    assert config.max_attempts == max_attempts
```
jet-tong, line 54: If we remove the upper limit guard, perhaps we can simply add test_max_attempts_one_succeeds, or use a smaller range (e.g. 1 to 10) for the test so upstream changes on the constant wouldn't fail our test silently.

jet-tong (changes requested):
Thanks for the PR - surfacing an appropriate error would always be better than the vague panic. I would suggest reshaping this PR so we fix the root cause of the panics instead of covering it up on the surface. 

1. Firstly, on `max_attempts` at 0: I believe the root cause fix for this would be in our Rust binding, where we used `.max_attempts(NonZeroUsize::try_from(max_attempts).expect("max_attempts must be > 0"));` instead of surfacing a python error like `PyValueError::new_err("max_attempts must be a positive integer")?` (as you also pointed out in comments). But I also think the post init guard is still nice to have on top to fail quickly since the client is built lazily, so I'd like to keep that part. (FYI I've also recommended in another PR to add a `__post_init__` guard seeing your PR how it's a good way to fail fast)
2. The upper bound - I'd recommend shipping only the lower-bound guard as you considered doing, here's why:
	1. The upper bound is actually off by one. CRT rejects `max_retries > 63`, but in [mountpoint-s3-client 0.14.1 s3_crt_client](https://github.com/awslabs/mountpoint-s3/blob/c6bc7dbc6a2982395dfc274045724d3710a4dbd5/mountpoint-s3-client/src/s3_crt_client.rs#L352) we set `max_retries = max_attempts.saturating_sub(1)` so ceiling is actually `max_attempts=64` and it only panics at 65 as the GitHub issue describes. 
	2. The full chain for this value is s3torchconnector -> s3torchconnectorclient -> mountpoint-s3-client -> ... -> aws-c-io. I would rather not hardcode a value owned by a dependency multiple layers upstream. It also doesn't feel right to test the CRT owned constant in our top python layer (which didn't catch the wrong upper bound in the PR). Perhaps we can fix this upstream in 2 steps, 1/ in awslabs/mountpoint-s3/mountpoint-s3-client s3_crt_client.rs `.unwrap()` -> `.map_err(NewClientError::CrtError)?` ([here](https://github.com/awslabs/mountpoint-s3/blob/c6bc7dbc6a2982395dfc274045724d3710a4dbd5/mountpoint-s3-client/src/s3_crt_client.rs#L355)), and 2/ awslabs/aws-c-io to surface a more appropriate error than `AWS_ERROR_INVALID_ARGUMENT` ([here](https://github.com/awslabs/aws-c-io/blob/54350963b64dfc6c4b0ea623b08aa252aae3d7d7/source/exponential_backoff_retry_strategy.c#L337-L341)).
	3. We could also consider to create an issue to aws-c-io to see if removing the upper limit would be viable, which is what issue #361 really wants to achieve. Until we get that confirmation I'd like to keep #361 open in the meantime.

These are changes I'd recommend for s3torchconnectorclient side in here to fix the 0 panic issue, expanding on point 1 above (I haven't tested them myself yet).

In `mountpoint_s3_client.rs`:
```rust
use pyo3::exceptions::PyValueError;

...

        let max_attempts_non_zero = NonZeroUsize::try_from(max_attempts)
	        .map_err(|_| PyValueError::new_err("max_attempts must be a positive integer"))?;

        let mut config = S3ClientConfig::new()
            .user_agent(UserAgent::new(Some(user_agent_string.to_owned())))
            .throughput_target_gbps(throughput_target_gbps)
            .part_size(part_size)
            .auth_config(auth_config)
            .endpoint_config(endpoint_config)
            .max_attempts(max_attempts_non_zero);
```

And a python test in `s3torchconnectorclient/python/tst/unit/test_mountpoint_s3_client.py` to test that 0 `max_attempts` doesn't panic and surfaces as a proper python error:
```python
def test_max_attempts_zero_raises():
    with pytest.raises(ValueError, match="max_attempts must be a positive integer"):
        MountpointS3Client(REGION, "", max_attempts=0)
```

In `test_s3_client_config` I'm happy to keep `test_max_attempts_negative_raises` and `test_max_attempts_zero_raises`; perhaps we can just check `test_max_attempts_one_succeeds` for the happy path. 

For `_MAX_ATTEMPTS_LIMIT` and the related post init / tests - let's drop them for this PR and see if we can fix it upstream first. I've recommended a soft documentation of the upper limit of 64 max_attempts - please have a look and see if that works for you.

jet-tong:
Hey @valter-silva-au, please let us know if you plan to update the PR.

PR #435, opened 2026-07-02
feat: add requester_pays option to S3ClientConfig

### LINKED ISSUE #430: expose Requester Pays (`x-amz-request-payer`) in `S3ClientConfig` / DCP `S3StorageWriter`/`S3StorageReader`
## Summary

`s3torchconnector` can't read from or write to **Requester Pays** S3 buckets, because there is no way to send the required `x-amz-request-payer: requester` header. The capability already exists in the underlying `mountpoint-s3` (its FUSE CLI has `--requester-pays`), but the `s3torchconnectorclient` Python binding doesn't surface it. Please add a `requester_pays` option to `S3ClientConfig` (and therefore to `MountpointS3Client`, the dataset/checkpoint APIs, and the DCP `S3StorageWriter`/`S3StorageReader`).

## Motivation

Some S3 buckets we need to access are configured as **Requester Pays**. Other AWS tooling (boto3, AWS CLI, s3fs) reads and writes these buckets fine by sending `RequestPayer=requester` on each request. `s3torchconnector` is currently the one client that can't, which blocks using it (e.g. its `dcp.S3StorageWriter`/`S3StorageReader` or the dataset APIs) against those buckets at all.

## Current behavior

Against a Requester Pays bucket, every operation fails with a bare `AccessDenied` (the request is missing `x-amz-request-payer`), and there is no Python-side way to inject the header:

```python
from s3torchconnectorclient._mountpoint_s3_client import MountpointS3Client

MountpointS3Client(region="us-east-1", requester_pays=True)
# TypeError: MountpointS3Client.__new__() got an unexpected keyword argument 'requester_pays'

c = MountpointS3Client(region="us-east-1")
c.put_object("bucket", "key", RequestPayer="requester")
# TypeError: MountpointS3Client.put_object() got an unexpected keyword argument 'RequestPayer'
```

`MountpointS3Client.__init__` (1.5.0 and `main`) exposes only:
`region, user_agent_prefix, throughput_target_gbps, part_size, profile, unsigned, endpoint, force_path_style, max_attempts` — no requester-pays option. The methods are compiled (pyo3), so they can't be monkeypatched to add the header from Python either.

## Proposed change

- Add `requester_pays: bool = False` to `S3ClientConfig` and to `MountpointS3Client` (mirroring the existing `unsigned` / `force_path_style` flags).
- When set, send `x-amz-request-payer: requester` on GET/PUT/HEAD/LIST/DELETE/multipart requests — the same behavior as `mountpoint-s3 --requester-pays`.
- Flow it through `S3ClientConfig` to the dataset APIs and the DCP `S3StorageWriter`/`S3StorageReader` (which already accept `s3client_config`).

Precedent: `mountpoint-s3` already implements this for its FUSE CLI (`--requester-pays`, documented under "Configuring and using Mountpoint"), so the native/CRT plumbing exists — this request is about surfacing it through the Python binding.

## Workaround and its cost

The only current alternative for a Requester Pays bucket is to drive `torch.distributed.checkpoint` through its `FsspecWriter`/`FsspecReader` backed by `s3fs` (which honors `requester_pays`). It works, but in a quick comparison it ran roughly **2× slower** than `s3torchconnector` on the same sharded checkpoint (≈155 vs ≈300 MB/s) — the s3fs/aiobotocore Python path doesn't match the CRT client. Exposing `requester_pays` would let users keep the faster CRT path.

## Environment

- `s3torchconnector` / `s3torchconnectorclient` 1.5.0
- Python 3.12, PyTorch 2.x, Linux x86_64

ph-tgs:
## Description

Adds a `requester_pays` option to `S3ClientConfig` so the connector can access [Requester Pays](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RequesterPaysBuckets.html) buckets.

When set, the `x-amz-request-payer: requester` header is sent on all S3 requests. The flag is threaded from `S3ClientConfig` through the pyo3 bindings to the underlying mountpoint-s3 client, where it is applied via `S3ClientConfig::request_payer("requester")` at the client level — so it covers GET/PUT/HEAD/LIST/DELETE and multipart requests without per-operation changes.

This resolves #430, which noted that users currently have to fall back to slower s3fs/`FsspecReader` alternatives (~2x slower) to read Requester Pays buckets.

## Additional context

- The implementation mirrors the existing `unsigned` flag: a new field on `S3ClientConfig`, plumbed through `_s3client.py` / `_mock_s3client.py` and the Rust `MountpointS3Client` / mock bindings (constructor signature, `#[pyo3(get)]` property, and pickle `__getnewargs__`).
- Applied at the client level rather than per-request, so no operation can accidentally omit the header.
- Exposed as a `bool` (matching `unsigned`) since `"requester"` is the only valid header value.
- **No breaking change.** `requester_pays` defaults to `False`; existing behavior is unchanged.
- The datasets, checkpoint, and DCP `S3StorageWriter`/`S3StorageReader` require no changes — they already accept and forward `S3ClientConfig`, so the new field flows through automatically.

## Related items

Closes Issue #430 but depends on #438 after merge to resolve. 

## Testing

- `cargo check` passes with the new `request_payer("requester")` call.
- Added unit tests:
  - `test_requester_pays_s3_client` / `test_requester_pays_s3_client_default` (s3torchconnector)
  - extended `test_mountpoint_client_pickles` to assert `requester_pays` survives a pickle round-trip (s3torchconnectorclient)
- Ran the full unit suites locally (built from source): `s3torchconnectorclient` unit suite passes (80 passed, 2 skipped) and the `s3torchconnector` unit suite passes aside from environment-only failures unrelated to this change (a user-agent test that assumes a non-RC Python, and DCP/Lightning modules requiring optional deps not installed in the local venv).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## TBD
+
+### New features
+* Add `requester_pays` option to `S3ClientConfig` to access Requester Pays buckets via the `x-amz-request-payer` header (#430)
+
 ## v1.5.0 (February 20, 2026)
 
 ### New features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## TBD
 
 ### New features
-* Add `requester_pays` option to `S3ClientConfig` to access Requester Pays buckets via the `x-amz-request-payer` header (#430)
+* Add `requester_pays` option to `S3ClientConfig` to access Requester Pays buckets via the `x-amz-request-payer` header (#435)
 
 ### Bug fixes
 
--- a/s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
+++ b/s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
@@ -20,6 +20,8 @@ class S3ClientConfig:
     force_path_style(bool): forceful path style addressing for S3 client.
     max_attempts(int): amount of retry attempts for retrieable errors.
     profile(str): Profile name to use for S3 authentication.
+    requester_pays(bool): Set to true to send the `x-amz-request-payer: requester` header on all
+        S3 requests, enabling access to Requester Pays buckets.
     """
 
     throughput_target_gbps: float = 10.0
@@ -28,3 +30,4 @@ class S3ClientConfig:
     force_path_style: bool = False
     max_attempts: int = 10
     profile: Optional[str] = None
+    requester_pays: bool = False
--- a/s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
+++ b/s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
@@ -21,7 +21,8 @@ class S3ClientConfig:
     max_attempts(int): amount of retry attempts for retrieable errors.
     profile(str): Profile name to use for S3 authentication.
     requester_pays(bool): Set to true to send the `x-amz-request-payer: requester` header on all
-        S3 requests, enabling access to Requester Pays buckets.
+        S3 requests, enabling access to Requester Pays buckets. Cannot be combined with
+        unsigned=True (requester pays requires authenticated requests).
     """
 
     throughput_target_gbps: float = 10.0
@@ -31,3 +32,10 @@ class S3ClientConfig:
     max_attempts: int = 10
     profile: Optional[str] = None
     requester_pays: bool = False
+
+    def __post_init__(self):
+        if self.requester_pays and self.unsigned:
+            raise ValueError(
+                "requester_pays=True requires authenticated requests and cannot be combined with unsigned=True. "
+                "For more information, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/RequesterPaysBuckets.html"
+            )
--- a/s3torchconnector/tst/unit/test_s3_client.py
+++ b/s3torchconnector/tst/unit/test_s3_client.py
@@ -168,6 +168,19 @@ def test_force_path_style_s3_client():
     assert s3_client._client.force_path_style is True
 
 
+def test_requester_pays_s3_client():
+    s3_client = S3Client(
+        region=TEST_REGION,
+        s3client_config=S3ClientConfig(requester_pays=True),
+    )
+    assert s3_client._client.requester_pays is True
+
+
+def test_requester_pays_s3_client_default():
+    s3_client = S3Client(region=TEST_REGION)
+    assert s3_client._client.requester_pays is False
+
+
 def test_s3_client_different_configs():
     s3_client_slow = S3Client(
         region=TEST_REGION,
--- a/s3torchconnector/tst/unit/test_s3_client.py
+++ b/s3torchconnector/tst/unit/test_s3_client.py
@@ -176,11 +176,6 @@ def test_requester_pays_s3_client():
     assert s3_client._client.requester_pays is True
 
 
-def test_requester_pays_s3_client_default():
-    s3_client = S3Client(region=TEST_REGION)
-    assert s3_client._client.requester_pays is False
-
-
 def test_s3_client_different_configs():
     s3_client_slow = S3Client(
         region=TEST_REGION,
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -153,6 +153,8 @@ Using S3ClientConfig you can set up the following parameters for the underlying 
 
 * `max_attempts(int)`: Number of retries for retriable errors
 
+* `requester_pays(bool)`: Set to true to send the `x-amz-request-payer: requester` header on all S3 requests, enabling access to Requester Pays buckets.
+
 For example, this can be passed in like: 
 ```py
 from s3torchconnector import S3MapDataset, S3ClientConfig
@@ -171,6 +173,10 @@ s3_lightning_checkpoint = S3LightningCheckpoint(region=REGION, s3client_config=c
 
 # Disable signing to make requests without AWS credentials
 config = S3ClientConfig(unsigned=True)
+s3_map_dataset = S3MapDataset.from_prefix(DATASET_URI, region=REGION, s3client_config=config)
+
+# Access a Requester Pays bucket
+config = S3ClientConfig(requester_pays=True)
 s3_map_dataset = S3MapDataset.from_prefix(DATASET_URI, region=REGION, s3client_config=config)
 ```
 
--- a/s3torchconnector/src/s3torchconnector/_s3client/_mock_s3client.py
+++ b/s3torchconnector/src/s3torchconnector/_s3client/_mock_s3client.py
@@ -40,6 +40,7 @@ class MockS3Client(S3Client):
             unsigned=self.s3client_config.unsigned,
             force_path_style=self.s3client_config.force_path_style,
             max_attempts=self.s3client_config.max_attempts,
+            requester_pays=self.s3client_config.requester_pays,
         )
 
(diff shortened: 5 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## TBD
 
 ### New features
+* Add `requester_pays` option to `S3ClientConfig` to access Requester Pays buckets via the `x-amz-request-payer` header (#430)
```
jet-tong: We usually use the PR numbers instead of issue numbers in here - would appreciate you switching number to #435.

s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
```diff
@@ -20,6 +20,8 @@ class S3ClientConfig:
     force_path_style(bool): forceful path style addressing for S3 client.
     max_attempts(int): amount of retry attempts for retrieable errors.
     profile(str): Profile name to use for S3 authentication.
+    requester_pays(bool): Set to true to send the `x-amz-request-payer: requester` header on all
+        S3 requests, enabling access to Requester Pays buckets.
```
jet-tong: Possible improvement: The [RequesterPaysBuckets doc](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RequesterPaysBuckets.html) has this note: "If you enable Requester Pays on a general purpose bucket, anonymous access to that bucket is not allowed.". I also tested this - unsigned requests would get 403 even with the header, and it feels right to reject the combination on our side instead of users getting an opaque 403 from S3. 

Could you please add a `__post_init__` to reject this combination early? e.g.:

```python
@dataclass(frozen=True)
class S3ClientConfig:
    """
    ...
    requester_pays(bool): Set to true to send the `x-amz-request-payer: requester` header on all
        S3 requests, enabling access to Requester Pays buckets. Cannot be combined with 
        unsigned=True (requester pays requires authenticated requests).
    """
    ...

    def __post_init__(self):
        if self.requester_pays and self.unsigned:
            raise ValueError(
                "requester_pays=True requires authenticated requests and cannot be combined with unsigned=True. "
                "For more information, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/RequesterPaysBuckets.html"
            )
```

s3torchconnector/tst/unit/test_s3_client.py
```diff
@@ -168,6 +168,19 @@ def test_force_path_style_s3_client():
...
+    s3_client = S3Client(
+        region=TEST_REGION,
+        s3client_config=S3ClientConfig(requester_pays=True),
+    )
+    assert s3_client._client.requester_pays is True
+
+
+def test_requester_pays_s3_client_default():
+    s3_client = S3Client(region=TEST_REGION)
+    assert s3_client._client.requester_pays is False
```
jet-tong: It'd be good to split tests into `test_s3_client_config.py` (for testing the `S3ClientConfig` dataclass) and `test_s3_client.py` (for testing the client itself) to follow existing style. 

Could you please remove `test_requester_pays_s3_client_default` from current `test_s3_client.py`, and modify tests to `test_s3_client_config.py`, e.g.:

```python
# Add requester_pays line to test_default()
def test_default():
    config = S3ClientConfig()
    assert config.part_size == 8 * MiB
    assert config.throughput_target_gbps == 10.0
    assert config.force_path_style is False
    assert config.max_attempts == 10
    assert config.profile is None
    assert config.requester_pays is False

...

# Before test_custom_setup, add:
def test_enable_requester_pays():
    config = S3ClientConfig(requester_pays=True)
    assert config.requester_pays is True
```

Would be good to add a test for unsigned + requester_pays rejection too in `test_s3_client_config.py`, if you took my suggestion above:
```python
# Need pytest import for pytest.raises
import pytest

# Before test_custom_setup
def test_requester_pays_with_unsigned_raises():
    with pytest.raises(ValueError):
        S3ClientConfig(requester_pays=True, unsigned=True)
```

jet-tong (commented):
Hey @ph-tgs, thanks for taking the time to create the PR! Your changes look great overall, I've just added some comments to suggest refactors / minor improvement opportunity.

I'd also like to add an integration test to ensure the feature stays tested against S3 in the future, but since it requires a bit of CI infra changes (and github workflow changes), we can handle that on our side as a follow-up PR instead.

CHANGELOG.md
ph-tgs: Fixed. Switched to the correct reference.

s3torchconnector/src/s3torchconnector/_s3client/s3client_config.py
ph-tgs: Added! Thanks for testing that and providing the snippet.

s3torchconnector/tst/unit/test_s3_client.py
ph-tgs: Done:

- Removed test_requester_pays_s3_client_default from test_s3_client.py
- Added requester_pays to test_default() in test_s3_client_config.py
- Added test_enable_requester_pays and test_requester_pays_with_unsigned_raises there

Kept test_requester_pays_s3_client in test_s3_client.py. It verifies client threading, mirroring test_unsigned_s3_client.

Thanks for the quick reviews!

ph-tgs:
Heads-up on a downstream gap this feature surfaces: with `requester_pays=True`, GET/PUT/HEAD work, but `S3FileSystem.rename` (copy + delete) fails with **403** on Requester Pays buckets.

Root cause is in aws-c-s3, not here: a `CopyObject` meta-request first issues a `HEAD` on the source object to size the copy, and that HEAD is built without the original request's headers — so `x-amz-request-payer` is dropped and the copy 403s before any other sub-request runs.

Tracked + fixed upstream:
- Issue: awslabs/aws-c-s3#652
- Fix PR: awslabs/aws-c-s3#654

Note the propagation path: aws-c-s3 → mountpoint-s3-crt-sys → mountpoint-s3 → this connector. `rename`/copy on Requester Pays buckets will keep 403ing until that fix lands here via a submodule bump. Until then the workaround is to write the final object directly with `PutObject` rather than tmp-write + rename-copy.

This doesn't block the client-config feature in this PR (the flag itself is correct) — flagging so the copy limitation is documented.

jet-tong (approved):
LGTM, thanks for addressing the comments.

Great catch on the CopyObject gap upstream! Note that aws-c-s3 + mountpoint-s3-client (in mountpoint-s3) releases are required before we can consume the changes so this fix might take a while. We should track this in a separate GitHub issue. 

Quick question: I'm aware only S3StorageWriter should be affected by the rename/CopyObject gap, but may I ask if writing a checkpoint with `requester_pays=True` is a common use case for you? (I was thinking usually only the account owner is responsible for writing the checkpoints so no `requester_pays=True` flag is required).

ph-tgs:
> LGTM, thanks for addressing the comments.
> 
> Great catch on the CopyObject gap upstream! Note that aws-c-s3 + mountpoint-s3-client (in mountpoint-s3) releases are required before we can consume the changes so this fix might take a while. We should track this in a separate GitHub issue.

No worries. I have my own release that our team is using for now. I'll open an issue that clearly links the PRs.

> Quick question: I'm aware only S3StorageWriter should be affected by the rename/CopyObject gap, but may I ask if writing a checkpoint with `requester_pays=True` is a common use case for you? (I was thinking usually only the account owner is responsible for writing the checkpoints so no `requester_pays=True` flag is required).

Its common for us but likely a somewhat unique case. 
We opted for a single self-managed MLFlow service across the company and attribute costs using `requester_pays=True` for this and other shared cross-account services.

